PR #102, opened 2015-01-10
Infrastructure layer improvements

Nakaner:
I have improved rendering of infrastructure view. Look at the commit messages for more details.

### DIFF
--- a/styles/standard.mapcss
+++ b/styles/standard.mapcss
@@ -78,6 +78,7 @@ way[tunnel=yes]::tunnels
 	opacity: 0.6;
 	linecap: butt;
 	linejoin: butt;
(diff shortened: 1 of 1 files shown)


rurseekatze:
The rendering of bridge:name does not work properly (tested in JOSM):

* the labels of bridge:name and ref collide; currently bridge:name is rendered on the top of ref
* the font size is too small to be readable; you should add some styling rules (I would prefer italic font style for a visible difference between bridge:name and name/ref):
```
text-position: line;
text-color: #585858;
font-size: 11;
font-family: Verdana Bold;
font-style: italic;
text-halo-radius: 1;
text-halo-color: white;
```

PR #125, opened 2015-02-12
Fix syntax errors and deprecation warnings

DerDakon:


rurseekatze:
I removed `fill-color` in https://github.com/rurseekatze/OpenRailwayMap/commit/41d7f1e959afdb02d4d8a27587a278ba2d24635a because this caused that tiles were rendered with a gray and opaque background. Now the background in JOSM is default black, but I don't think that this is a problem for the users.

Issue #95, opened 2015-01-08
Render level crossings in infrastructure layer


Nakaner:
Level crossings should be rendered in infrastructure layer because they are a common reason why a train stops outside stations.

PR #134, opened 2015-02-17
Crossing icons

Nakaner:
This pull request solves #95. The black and white icons are used if there is only few information about the crossing.


Nakaner:
@rurseekatze Could please merge this pull request or shall I draw a third version of the icons?

rurseekatze:
I think that the icons for level crossings are ok (although the German saltire is not the best solution), but I do not like the crossing icons yet.

The crossing icons look like little country flags or something like that, but nobody would know what they mean without having a look at the legend. I think that icons should be so simple, that every user roughly knows what it symbolizes without any explaination. The crossing icons are a bit too abstract in my point of view.

PR #153, opened 2015-02-22
change errors to warnings if the error is not severe

### LINKED ISSUE #135: Vaildator: warnings and erros (priority=* etc.)
Following tags should generate a warning instead of an "error":
* priority=*
* signal tags without DE-ESO: prefix

The first one does not harm if it is left at an way because its damage is much lower than a invalid multipolygon. The second one should generate a warning as long as ORM renders the old tags.

Nakaner:
fix #135


DerDakon:
* comments never match warning texts
* fixAdd on first rule is wrong, the tag is already present

rurseekatze:
As @DerDakon said, there are still some mistakes. Please fix them before, so that I can merge this pull request.

Nakaner:
Is it ok now?

PR #154, opened 2015-02-22
add rendering of abandoned:name instead of name=* on railway=abandoned

### LINKED ISSUE #114: Use abandoned:name to render the name of razed tracks.
This is a suggestion first and foremost. 

There are instances where a track was demolished and a new feature was built upon it. Take [this example](http://www.openstreetmap.org/way/325552701) where a cycleway follows the razed railway line. 

What happens is that the name of the new feature appears on the renderer [instead of the name of the track that was razed](http://www.openrailwaymap.org/?lang=en&lat=42.0291740276681&lon=-8.635864108800888&zoom=19&style=standard).

There's also the tag [old_name](http://taginfo.openstreetmap.org/keys/old_name#overview) which is a lot more used than [abandoned:name](http://taginfo.openstreetmap.org/keys/abandoned%3Aname#overview).

Nakaner:
This fixes #114

### DIFF
--- a/styles/standard.mapcss
+++ b/styles/standard.mapcss
@@ -120,8 +120,8 @@ way[tunnel=yes]::tunnels
 }
 
 /******************************************************************************/
-/* refs and names                                                             */
-/* We have to exclude bridge and tunnels in order to prevent a label collison */
+/* text of tracks outside bridges and tunnels                                 */
+/* default track label contains ref and name                                  */
 /******************************************************************************/
 way[railway=rail][!service][!bridge][!tunnel],
 way[railway=disused][!service][!bridge][!tunnel],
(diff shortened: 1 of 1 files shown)


DerDakon:
* in case abandoned:ref is preset this should be preferred over ref, too
* this probably should also be applied to razed tracks?
* the third hunk only adds whitespace noise

Nakaner:
@DerDakon I added support for abandoned:ref=*, razed:name=* and razed:ref=*.

DerDakon:
There is a trailing whitespace in the first hunk.

Remove, then git commit --amend styles/standard.mapcss && git push --force

DerDakon:
To fix the whitespace issue in the first commit:

git checkout issue-114-abandoned-name
git rebase -i HEAD~2 ## change the "pick" in the first line to "edit", then save
echo -e 'n\nn\ny' | git checkout --patch issue-114-abandoned-name styles/standard.mapcss
git rebase --continue
git push --force

I wonder if it isn't much easier to drop everything but the "text" line from all but the first statements: it's only the text line that is changed, and repeating everything else makes this harder to find out.

Nakaner:
I fixed all the issues @DerDakon mentioned. This PR can be merged now.

DerDakon:
* you have several instances of eval(tag("x")), which can be changed to a simple "x"
* I would keep the default text when defining the rail rendering, and just overwrite it later, this saves one markup block

And I would suggest squashing everything together, as you have many changes that live only for one commit and are then overwritten.

PR #157, opened 2015-02-22
add error notifications if usage=freight is used

### LINKED ISSUE #155: Validator should notify about the usage of usage=freight
Validator should notify about the usage of usage=freight as an error and offer railway:traffic_mode=freight instead.

Nakaner:
This fixes #155


DerDakon:
* the comment on the first validator rule does not match the warning
* the assertNoMatch lines do not have "way" specified
* I doubt the assertMatch on the second rule matches (hint: validator.check_assert_local_rules must be "true" in JOSM)

PR #161, opened 2015-02-24
validator: allow usage+service for usage=industrial

DerDakon:


rurseekatze:
`usage=*` and service=*` should also be allowed for `usage=military`. Please add that before I will merge your pull request.

PR #163, opened 2015-02-28
render track refs when usage is given

DerDakon:
The layout is similar to the signs used in stations in Germany by Deutsche
Bahn. This makes them much more prominent on higher zoom levels. On zoom levels
below 16 the track refs are not drawn anymore.


DerDakon:
Fixes issue #164 .

DerDakon:
I now also added "shield" rendering rules, although most implementations do not support that (yet), but for the case they eventually will this will hopefully give a nicer result.

Issue #124, opened 2015-02-10
Electrification layer


rurseekatze:
Add a separate layer which visualizes electrification of railroads. It should consider the following tags:
- `electrified=yes/no/contact_line/rail/ground-level_power_supply`
- `voltage=*`
- `frequency=*`
- `electrified:rail=top/side/bottom`
- `railway:signal:electricity:*=*`
- `deelectrified=yes/contact_line/rail`
- `railway=isolated_track_section`
- `railway:main_switch_off=yes`
- `railway:lower_pantograph_section=yes`
- `railway=power_supply`
- `railway:electricity=joint` and `railway:electricity:jumpering=jumpered/possible/no`

JJJWegdam:
**Proposal for rendering traction systems**
In this style, railways have coloured through their voltage/frequency according to the world's standard systems [[1]](http://en.wikipedia.org/wiki/List_of_current_systems_for_electric_rail_traction). The downside is that other systems [[2]](
http://en.wikipedia.org/wiki/List_of_current_systems_for_electric_rail_traction#Systems_using_non-standard_voltages) will be displayed in black. 
Another method would be to use smaller color intervals in order to cover the whole spectrum of voltages. Downsides would be that small differences (600 V vs 750 V within Rotterdam for example) might get lost and that different frequencies can't be mapped at all (which is a problem at 25 kV). For these reasons I prefer the method described below.

![Alt Text](http://puu.sh/gYIGq/dbdd25dc5d.png)


**Proposal for rendering Dutch electrical signs**
The Netherlands has some specific signs for traction like described in an ORM wiki [section](https://wiki.openstreetmap.org/wiki/OpenRailwayMap/Tagging_in_Netherlands#Pantograph_directions). I'd propose to render 306a (turn off the main switch) at high zoom levels; this because it is (in the Netherlands) present in all combinations of electrical signs. For example: at several Dutch moveable bridges a piece of catenary is not present. Before the train driver is asked to lower the pantograph, allways a sign with turning off the main switch is displayed. This to prevent an electric arc (which damages the catenaries and pantograph).
Voltage switches and neutral sections are simply displayed by the tracks and other signs are allways in the vicinity of a 306a. Therefore I would advice to render these signs only at lower zoom levels.

![Alt Text](https://wiki.openstreetmap.org/w/images/thumb/e/ea/Dutch_railway_traction_signs.png/1024px-Dutch_railway_traction_signs.png)

[...]

PR #172, opened 2015-03-23
Issue 171 catenary signals

### LINKED ISSUE #171: Add German/Austrian catenary signals
Add following German catenary signals (El). They are also used in Austria but have different names there:

* El 1
* El 1v
* El 2
* El 3
* El 4
* El 5
* El 6
* direction arrow used on El signals

Nakaner:
This pull requests adds
* German signals El 1, El 1v, El 2, … El 6
* right and left direction arrows used on El 6 signs
* Austrian equivalents of the signs called above

This fixes #171


JJJWegdam:
I propose to merge this with [#124](https://github.com/rurseekatze/OpenRailwayMap/issues/124).

Nakaner:
I agree @DerDakon that we should render catenary signals only at the electrification layer (canvas-based map only) to reduce the "noise" at signal layer.

@DerDakon You can move the whole code to the branch electricity-layer in your fork and create a pull request afterwards. As @rurseekatze said, it is not possible to serve it as raster tiles (due to hardware restrictions) but the layer can be used for mapping in JOSM and at our [canvas variant](http://www.openrailwaymap.org/canvas.php).

DerDakon:
I have updated my branch. Please check if I have transferred everything correctly. Afterwards remove the changes in signals.mapcss from your branch and squash everything together.

Nakaner:
I hereby close this pull request because it has been replaced by pull request #221 which fixes the same issue in a better way.

PR #181, opened 2015-04-02
Validator rules warning if VzG and KBS numbers are tagged the wrong way

### LINKED ISSUE #152: Validator rules for German tagging issues
Add validator rules to detect some German tagging issues:

* `name` or `ref` contains `KBS`
* `name` or `ref` contains `Kursbuchstrecke`
* `name` or `ref` contains `VzG`
* `name` or `ref` contains `KBS`
* `ref` contains the track number, e.g. `ref=2610-2` instead of `ref=2610` and `railway:track_ref=2`

Nakaner:
This fixes #152.

This pull request add validator rules looking for KBS and VzG numbers.

### DIFF
--- a/validator/de-openrailwaymap.validator.mapcss
+++ b/validator/de-openrailwaymap.validator.mapcss
@@ -1,0 +1,122 @@
+meta
+{
+	title: "OpenRailwayMap – extra ruleset for Germany";
+	version: "2015-05-26";
+	description: "Additional validation of railway data according to the OpenRailwayMap tagging scheme.";
+	author: "rurseekatze";
+	link: "https://wiki.openstreetmap.org/wiki/OpenRailwayMap/Tagging";
+	watch-modified: true;
+}
+
+/* track names containing "KBS" */
+way[railway][name=~/^KBS [0-9]*.*/],
+way[railway][name=~/^Kursbuchstrecke [0-9]*.*/]
+{
+	throwError: "Tracks should not be named by their timetable number (KBS xy). Use a route relation with route=railway, instead.";
+	assertMatch: "way railway=rail name=\"KBS 258\"";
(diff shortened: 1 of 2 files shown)


DerDakon:
-please drop the first part, which is the same as #210
-"way[railway][name=~/^VzG [0-9]*.*/]" is there twice
-KBS numbers could look like "360.1.2", so the last regexes should be ~/^[0-9]{3}\.[0-9]{1,2}[-.][0-9]{1,2}$/] (please also add a positive and negative test to make sure that the dot isn't takes as wildcard at this point)

Nakaner:
@DerDakon I have fixed everything you mentioned.

DerDakon:
The positive test for 2 dots in KBS numbers is missing:

assertMatch: "way railway=rail ref=\"790.4.5\"";
assertMatch: "way railway=rail name=\"790.4.5\"";

Also the last test is about ref and name, but only mentions name, which can be misleading.

And I would squash everything into one commit, you have a lot of changes that only live for one commit and then are replaced by something different.

Nakaner:
@DerDakon wrote:
> There is "VzG" in 3 error texts, which probably must be \"VzG\" instead

There is no need to fix that. I only use quotation marks at the error texts where VzG is meant as part of the tag value.

I squashed my commits and added some missing tests.

DerDakon:
railway rail with name or ref of 3 digits may also be a wrong railway:track_ref. I have actually seen instances of that error, so we should give a different error message for this case.

Nakaner:
@DerDakon I have fixed this.

DerDakon:
You added that for name, but not for ref.

DerDakon:
IMHO you should only show the track_ref message for ref and name that matches [0-9]{3}, not for those that have additional characters.

Nakaner:
@DerDakon I have addressed your comments.

PR #201, opened 2015-04-21
Issue 182 german zs3 signals

### LINKED ISSUE #182: Render Zs 3 light signals
Render Zs 3 light signals. The the first character listed in `railway:signal:speed_limit:speed=*` should be rendered.

Nakaner:
This fixes #182 

I decided to use the signals from Bundesbahn because they have less light dots, i.e. drawing them is easier in Inkscape and they are better for low density icons.


DerDakon:
The light signals at Hannover Flughafen do only show one value, and will always show this value when the main signals are not Hp0 (see http://www.eisenbahn-unfalluntersuchung.de/SharedDocs/Publikationen/EUB/DE/Untersuchungsberichte/aelter/004_Hannover-Langenhagen.html). This will not work with your regex.

I would propose changing it to e.g. (untested) "/^120(;[0-9]+)*(;off)?$/"

DerDakon:
"?" is another valid part of that list.

DerDakon:
Looks like JOSM doesn't like to match the question mark in a reqular expression, regardless of how many quoting I add. So the only regex that currently works is =~/^120(;.*)?$/

Nakaner:
This pull request can be merged now.

DerDakon:
There is a fixed JOSM bug: https://josm.openstreetmap.de/ticket/11598, so one needs at least JOSM 8535 for quotation mark matching to work. This should also get a validator check, so everything that is not rendered because of a bad value is flagged as warning in JOSM.

DerDakon:
You should probably borrow some ideas from #282 to reduce the code size.

Nakaner:
I have resolved all conflicts, this PR can be merged. Code simplification will be done as an separate PR to get this PR merged soon.

Nakaner:
I close this PR because these changes will be introduced by another PR with another base.

PR #202, opened 2015-04-21
infrastructure layer: render turntables

DerDakon:


rurseekatze:
Could you also add railway=traverser?

DerDakon:
Done

Issue #40, opened 2014-06-02
Infrastructure Mode: render usage=tourism and usage=freight


Nzara:

Nakaner:
I think that usage=freight should not be rendered because it should be tagged railway:traffic_mode=freight.

Nakaner:
As far as I know, we deprecated usage=tourism at Cologne Meeting in July 2014 because of problems to define it.

@rurseekatze If you agree me, you can close this issue.

almich:
If a way with usage=freight or usage=tourism is not rendered with a special style the presence of a usage tag should at least not prevent the renderer from rendering. At the moment you can hide a line on openrailwaymap by setting usage=tourism for example. See the Brockenbahn in Germany for example: http://www.openrailwaymap.org/?lang=de&lat=51.76050924689838&lon=10.707035064697266&zoom=12&style=standard . By the way, usage=tourism is used at the moment on 5172 ways worldwide.

rurseekatze:
I think that usage=tourism should not be deprecated. Otherwise there would not be any suitable tag for tourism-only railways (like Selfkantbahn or Drachenfelsbahn).

Our next steps should be:
* first render tracks tagged with `usage=tourism/military/test` in black to make them visible on the map
* in the second step, think of suitable colours for `tourism`, `military` and `test`

Nakaner:
Step 1 ("render tracks tagged with `usage=tourism/military/test` in black to make them visible on the map") has been implemented by Dakon in pull request #207.

rurseekatze:
Tracks tagged with `usage=tourism/military/test` are now rendered in black.

We should collect ideas for suitable colours.

RavuAlHemio:
Hi! Tracks tagged with `usage=tourism` apparently aren't rendered anymore; taking the Kaltenleutgebner Bahn in the southwest of Vienna as an example, while [way 378286407](https://www.openstreetmap.org/way/378286407) appears as a thin black line, [way 780473885](https://www.openstreetmap.org/way/780473885) does not appear at all, and the difference is in the `usage=tourism` tag.

(The Kaltenleutgebner Bahn is a preserved railway only used for tourism, but there are intentions by the governments of Vienna and Lower Austria to reactivate it as light rail in the coming years.)

DerDakon:
See OpenRailwayMap/OpenRailwayMap-CartoCSS#7

twatzl:
Are there any news about this?
[...]

PR #207, opened 2015-04-22
Render also railway lines with usage tagged as tourism, military, or test

DerDakon:
Should fix issue #40.

### DIFF
--- a/styles/standard.mapcss
+++ b/styles/standard.mapcss
@@ -15,7 +15,7 @@ Format details:
 meta
 {
 	title: "OpenRailwayMap infrastructure standard layer";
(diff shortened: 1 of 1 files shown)


Nakaner:
I think that we as the renderer/map should not encourage users to use `usage=freight`, i.e. you should replace following line

    way|z9-[railway=narrow_gauge][!service]

by

    way|z9-[railway=narrow_gauge][!service][usage=tourism],
    way|z9-[railway=narrow_gauge][!service][usage=military],
    way|z9-[railway=narrow_gauge][!service][usage=test]

DerDakon:
Done

Issue #192, opened 2015-04-12
Add some signal related validation rules


Nakaner:
Add following validation rules to the validator:

* Signal numbers should be tagged `ref=*`, not `railway:ref=*`.
* A node may not be tagged both with `railway:signal:main=DE-ESO:ks` and `railway:signal:distant=DE-ESO:ks`.
* A node may not be tagged both with `railway:signal:main=*` and `railway:signal:combined=DE-ESO:ks`.
* A node may not be tagged both with `railway:signal:distant=*` and `railway:signal:combined=DE-ESO:ks`.
* Signals usually have no `name=*`.
* Signals should have a `railway:signal:direction=*` and `railway:signal:position=*`.
* Switches usually have no `name=*`. 

Some of this rules check errors JOSM user don't do. But if a JOSM user edits an way (even by inserting an additional node/signal), the validator checks all nodes of the way and reports bugs. This might let some mappers check for other tagging bugs in this area, too.

I have found all these bugs while I reviewed a lot of signals in Germany during [a mechanical edit](https://wiki.openstreetmap.org/wiki/User:Nakaner/Mechanical_Edit_German_on_Railway_Signals).

Nakaner:
* A stop sign (Ne 5) should not share a node with `public_transport=stop_position`.

DerDakon:
The Ne5 one isn't always true (see Lemmie/HLEM, northern tracks), e.g. when the Ne5 are set up that way that the end of the train is always at the same position, the track is usually operated with trains of length 2*N: the Ne5 for length N is at the same point as the middle of the usual train length -> both are at the same place.

All other warnings are already present.

PR #209, opened 2015-04-23
Signals without direction

### LINKED ISSUE #195: Do not render signals without railway:direction=*
A signal which does not have railway:direction is a information with few value. I suggest not to render theses signals anymore to force iD users to enter this tag.

Nakaner:
This fixes #195. Signals without direction won't be rendered anymore (both in maxspeed and signal layer).

This implements a part of #192.

### DIFF
--- a/styles/maxspeed.mapcss
+++ b/styles/maxspeed.mapcss
@@ -895,8 +895,8 @@ way|z10-["construction:railway=rail"][maxspeed>380][maxspeed<=400]
 }
 
 /* German speed signals (Zs 10) */
-node|z16-[railway=signal]["railway:signal:speed_limit"="DE-ESO:db:zs10"]["railway:signal:speed_limit:form"=sign]["railway:signal:speed_limit:speed"=none],
-node|z16-[railway=signal]["railway:signal:speed_limit"="db:zs10"]["railway:signal:speed_limit:form"=sign]["railway:signal:speed_limit:speed"=none]
+node|z16-[railway=signal]["railway:signal:direction"]["railway:signal:speed_limit"="DE-ESO:db:zs10"]["railway:signal:speed_limit:form"=sign]["railway:signal:speed_limit:speed"=none],
+node|z16-[railway=signal]["railway:signal:direction"]["railway:signal:speed_limit"="db:zs10"]["railway:signal:speed_limit:form"=sign]["railway:signal:speed_limit:speed"=none]
 {
 	z-index: 90;
 	icon-image: "icons/de-zs10-sign-44.png";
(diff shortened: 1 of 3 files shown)


bigbug21:
I oppose this change. There is a still some value in such incompletely tagged signals. In fact, there is no apparent value in those direction tags so far, although there might be some later (for example for railway simulators).

Nakaner:
@bigbug21 There are users at OSM (e.g. PeterDRS) who do not add `railway:direction=*` to their signals. We want to "force" users to enter well tagged signals and use our power as renderer to do so. PeterDRS is not the only user we want to guide using our power as renderer.

I may soon publish an additional style sheet for editing purposes (not for rendering on a server) which shows a danger icon (like JOSM does if you tag highway=residential on a node) if tagging of an object is broken.

rurseekatze:
I agree with Nakaner. Sure, the direction tag is currently not used by any application. But the direction tag provides essential information about the signal, and signals without this tag are not complete. So we should use our possibilities to force mappers to create valid data.
Just marking those signals in a validator does not have the same impact because many mappers do not pay attention to tagging issue warnings. But they pay attention to data that is not rendered on the map because most mappers tag for the renderer (although it should be opposite).

rurseekatze:
@Nakaner At some lines you forgot to add the selectors. I added them in https://github.com/rurseekatze/OpenRailwayMap/commit/7f3833fd423801a74edfb94c08f90811d0510a5a.

bigbug21:
For once, there are some signals (such as 'LZB-Bereichskennzeichen') that usually have two direcitons but cannot be edited accordingly by the JOSM template (yet). The moment this patch goes live, we will see most of these signals disappear for no reason.

Also, this will raise the already quite high bar for new signal mappers. There are mappers that have a hard time even uploading a single signal for a start. New mappers that are proud to see that single signal. They might be discouraged by unnecessary bars that prevent there first signal works from appearing on the map. We have come a long way to lower the bar of editing railway details by providing comprehensive templates and tutorials. Now we raise the bar again, for no reason.

There surely is a long way to go in terms of quality assessment and quality assurance. Better validator rules and a straight approach to errors (and their human causes ;)) are surely necessary. We have done a quite good job, I guess, in encouraging a small crowd of people to map railway details. Even in well-mapped Germany we still lack most railway details. Perfection should (and must) come later.

If we continue on this path,we will gradually exclude more and more signals from the map. So what's next? Will signals without a reference tag disappear? Or those without signal states? Will there eventually be an OpenRailwayMap where only a perfectly mapped signals will appear? Well, that might be the first of its kind within the OpenStreetMap universe.

A well-tagged signal without a direction tag is still a signal worth showing on the map. It still provides key information. If there was a direction tag attached to it, it would still appear in the very same way. There is no difference at all.

PR #210, opened 2015-04-23
Fix validator syntax

Nakaner:
"area" is an illegal object type. I replaced it by "relation".

This removes some exceptions JOSM throws.

### DIFF
--- a/validator/openrailwaymap.validator.mapcss
+++ b/validator/openrailwaymap.validator.mapcss
@@ -20,12 +20,12 @@ way[railway][usage][usage!=industrial][usage!=military][service]
 	assertNoMatch: "way railway=rail usage=military service=yard";
 }
 
-/* Station mapped as way or area */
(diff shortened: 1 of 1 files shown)


rurseekatze:
Why did you remove `area`? It is a valid selector: https://josm.openstreetmap.de/wiki/Help/Styles/MapCSSImplementation#Selectors

I would suggest to keep `area` and add `relation`.

DerDakon:
The errors come only from the "area" in asserts, the actual selectors are fine.

DerDakon:
Just drop this one and update your JOSM: https://josm.openstreetmap.de/ticket/11370#comment:3

rurseekatze:
Issue can be closed because there is nothing to do on our side.

I will add the check for `railway=station/crossing/level_crossing` mapped as relations in a separate commit.

rurseekatze:
Nothing to do on our side, issue can be closed.

PR #219, opened 2015-05-10
Render bridges and tunnels only if their railway lines are also rendered

### LINKED ISSUE #214: Tram bridge render bug
Tram bridges keep rendering after zoom 10, while their respective tram lines stop rendering at those zoom levels.

Alt text

DerDakon:
This prevents the bridges and tunnels from being rendered if the railway line is not rendered, e.g. tram lines on lower zoom levels.

Should fix #214.

### DIFF
--- a/styles/standard.mapcss
+++ b/styles/standard.mapcss
@@ -93,7 +93,17 @@ way|z9-[railway=razed]
 /*************************************************************/
 /* tunnel label (not render if the track has a track number) */
 /*************************************************************/
-way[tunnel=yes][!"railway:track_ref"]::tunnels
+way|z9-[railway=rail][tunnel=yes][!"railway:track_ref"]::tunnels,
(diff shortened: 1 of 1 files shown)


styles/standard.mapcss
```diff
@@ -93,7 +93,10 @@ way|z9-[railway=razed]
 /*************************************************************/
 /* tunnel label (not render if the track has a track number) */
 /*************************************************************/
-way[tunnel=yes][!"railway:track_ref"]::tunnels
+way|z9-[tunnel=yes][!"railway:track_ref"]::tunnels,
```
rurseekatze: I think you should add railway=rail/disused/abandoned/... here, otherwise e.g. tram tracks would also match this condition (which they should not do).

DerDakon:
Done.

PR #223, opened 2015-05-18
validator: only warn about duplicate operator if it really is duplicate

DerDakon:
Do not flag this tagging as error if the railway line has no operator set.


DerDakon:
This is currently broken because of https://josm.openstreetmap.de/ticket/11568#ticket .

DerDakon:
I now used a different operator, which seems to work reliable.

PR #226, opened 2015-05-27
style: add color to proposed and constructed railways

DerDakon:
The same color as for the operational lines is used.


Nakaner:
I think that using the same colour as for operational lines is no good idea. At the moment dashed lines do not look like dashed lines because ways are often splitted at OSM due to bridges and tunnels on the line. We should fix the rendering of dashed lines at low zooms before deploying this patch.

DerDakon:
This tends to happen far less for proposed or construction railways, as the reason for splitting is often things like speed change, PT relations at switches, and stuff like this. This usually does not happen until a railway line is constructed.

The place where it is most obvious IMHO that the rendering has a problem is SFS Ebensfeld-Erfurt, but the problems there seem to happen because the 2 (or more) parallel tracks overlap so something like a solid line seems to appear.

DerDakon:
We could shorten the dashes, so there is more gap than dash. That would make things better even for 2 rail tracks.

DerDakon:
This is based on #374 now, so please merge that one first.

PR #244, opened 2015-06-26
render caption given for German Ne5 signals

DerDakon:


PeterDRS:
I support this pull request. An Ne 5 without the caption is senseless.

DerDakon:
Some of them do not have a caption, usually those at the end of a platform.

PeterDRS:
That's no problem, just don't render any text in this case.

PeterDRS:
(I mean a single Ne 5 without a caption is senseless, having a bunch of Ne 5 without captions is also senseless)

DerDakon:
There are stations where you have only a single Ne5 for one direction, e.g. when that is the direction of the only way to the platform. See halt Lemmie for an example: the southern track has only a single Ne5, which is at the end of the platform and has no caption.

So, yes, it is weird if you have multiple Ne5 and none of them has a caption. The other option may be a valid case.

PeterDRS:
I think this is a github pull request for adding a text to Ne 5 having a caption.

Ne 5 without a caption stay untouched. Where is the problem? Is there any problem???

rurseekatze:
I do not understand this confusing discussion... What is the problem?

If there is a caption, then it is rendered, if there is no caption, then there will be just an icon rendered on the map?!

PeterDRS:
The problem is, that the pull request is not merged yet. Why?

Nakaner:
> The problem is, that the pull request is not merged yet. Why?

Please fork the project and set up your own server and compete with us for donors and their donations. Please do not ask useless questions which just waste the rare time the maintainer has.

This post is a wast of characters, time and space. ;-)

PeterDRS:
Why is it not allowed to ask why it is not merged?

When mapping Ne 5 it would be nice to see the result.

Apart from this, you think I am a person that wants to GET everything from others an does not GIVE anything.

That is fully wrong. I add much data to the OSM database. Without that data your server would show nothing. Always think of that.

rurseekatze:
Calm down, everbody! I did not expect that this simple pull requests results in such a long discussion...

> The problem is, that the pull request is not merged yet. Why?
> Why is it not allowed to ask why it is not merged?

The question is not why it is not merged yet, but when it will be merged. I am a busy person doing this stuff in my freetime without any payment. So it can take some time until a pull request gets merged.

> Apart from this, you think I am a person that wants to GET everything from others an does not GIVE anything.

I do not think that. But I do not like it when users put me under pressure. Or would you like it to get asked why you have not mapped the signals in station xy yet?

I hope that this nonsense discussion of offtopic stuff will end soon...

PeterDRS:
I do not know if it is complicated to merge this or not. But when tagging lots of Ne 5 it is annoying when they are not shown.

So maybe there is a reason for example a technical reason why it is not merged. That could be solved. If it is just time it is ok but should be written so that one can understand it.

DerDakon:
This can not be merged yet, as it uses cond(), which currently doesn't work in KothicJS.

DerDakon:
In case you want to merge this before Kothic is improved, I have prepared a workaround that I can push.

PR #245, opened 2015-06-26
presets: drop company type from examples

DerDakon:
http://wiki.openstreetmap.org/wiki/DE:Key:operator suggests to leave out the company form.


Nakaner:
There is no consensus inside OSM community on this topic. I would not merge this PR.

Issue #247, opened 2015-06-28
railway:signal:*:states cannot be selected


Nakaner:
Some Multiselects at our JOSM presets cannot be selected due to a `default=""`. That's why the multiselect where you can enter possible signal states (e.g. Hp 0, Hp 1, Hp 2) gray.

rurseekatze:
Was fixed by #248.

PR #248, opened 2015-06-28
Remove default="" from all multiselect

Nakaner:
Now you can add `railway:signal:main:states=*` from the preset without opening the preset twice.


Nakaner:
This fixes #247

PR #249, opened 2015-06-28
Add an additional style for JOSM highlighting some additional things

Nakaner:
This style renders bad tagged signals and objects which have no icon at web rendering. deprecated.png is from JOSM SVN `trunk/images/styles/standard/misc/deprecated.png`.


Nakaner:
Why did you add the label "Website"?

PR #251, opened 2015-07-05
Improve validator warnings about obsolete tagging

DerDakon:
- power:type is almost gone, raise message level to error
 - warn about radio=*, railway:radio=* should be used instead


rurseekatze:
The second validation rule you added is not correct. It adds `railway:radio=gsm-r` even if a track is not tagged with `radio=GSM-R` (e.g. `radio=analogue`).

I think it should be correct if you remove the `fixAdd` and `fixRemove`.

PR #255, opened 2015-07-14
show deactivation cross only for those signals that are rendered on the given layer

DerDakon:
This avoids stale deactivation crosses on the signal layer.

### DIFF
--- a/styles/maxspeed.mapcss
+++ b/styles/maxspeed.mapcss
@@ -28,6 +28,17 @@ canvas
 	default-lines: true;
 }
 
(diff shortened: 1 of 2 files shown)


rurseekatze:
Why should deactivation crosses not be rendered for minor and crossing signals? Those signals are already rendered in signalling layer.

What was your intention when using two comment beginnings (lines 1584 and 1585)?

DerDakon:
It was enabled for minor signals. That is was missing for crossing signals was just an oversight.

Sadly this will still only show the deactivation crosses, as they are at the same node as the signal, and only the last rule for a given node will be used.

PR #256, opened 2015-07-19
validator: allow signs to have a single state

DerDakon:
Until now the validator warned if anything was given in *:states for a signal
that has *:form=sign. However there instances of a signal with only one state
which is then given as only a sign (e. g. the German Bü0 signal), so only warn
if multiple states are given.


Nakaner:
Bü signals which cannot show Bü 1 should be tagged as signs, not light signals because they have no light. There are two signals of this type in Bad Nauheim. See also signals.mapcss how they are rendered.

I think that parts of this pull request should not be merged.

DerDakon:
I would definitely allow (if not require) also the single state that a sign could have to be put into the tagging. The railway tagging is verbose at many points, so I don't see in a value to explicitely forbid to put that information here.

Also the current restriction is probably again very much German-centric, I'm pretty sure that other countries could allow different single-state signals (something like either "hp0" or "hp2" as sign), so one then must explicitely tag which of the different values are shown at a given signal.

PR #257, opened 2015-07-19
presets: fix Zs3 and Zs3v state delimiter

DerDakon:
If multiple states were given they were separated by ',' instead of ';'.

This was introduced in commit 17803c17b899301d2aa9cdc7065288d199bfd1b3
(correct labels, change some comboboxes to multiselect) as before that only a
single value could have been entered by the presets.


Nakaner:
I suggest to merge this pull request as soon as possible.

PR #259, opened 2015-07-19
Some updates to zs2, zs2v, zs3

bigbug21:
zs2/zs2v cannot be fixed signs; zs3 might also be "off"/"dark" (note: There are speed signals that are not multiples of 10, such as "0.5" [5 kph]) which cannot be modelled through the current version of the template anymore)


Nakaner:
If you remove `<combo key="railway:signal:route:form" …`, you have to add `<key key="railway:signal:route:form" value="light">`. Otherwise the preset does not add `railway:signal:route:form=light` anymore! The same applies for route distant signals.

DerDakon:
You probably need to add "off" to Zs3v, too.

And it's probably a wise idea to rebase your topic on PR #257 as both touch the same lines and that one will probably get merged earlier.

DerDakon:
#257 is now in master, so please rebase on that.

bigbug21:
Thanks for your input! I have filed a new pull request at #263

DerDakon:
You could have just force-pushed to this branch.

PR #260, opened 2015-07-24
signals: show German Ks signals if they have additional states

DerDakon:
Allow either "off" or "DE-ESO:kennlicht" as additional states.


DerDakon:
Either merge this one or #273.

DerDakon:
This only works in JOSM if you have at least version 8535 because of [JOSM bug 11598](https://josm.openstreetmap.de/ticket/11598).

rurseekatze:
I think that the problem also occurs at Ks combined signals. Can you also fix that issue for these signals?

DerDakon:
Done for both combined and distant signals, and it allows now multiple of the additional states to be present.

PR #261, opened 2015-07-25
increase the spacing for abandoned lines

DerDakon:
When the line is a corner the segments will come close to each other so they
appear as a continuous line. I suspect that will get better if the dashes are
a bit farther apart.


DerDakon:
abandoned-join

example of problem: Süntelbahn http://www.openrailwaymap.org/canvas.php?lang=de&lat=52.197007066922616&lon=9.450516700744629&zoom=14&style=standard

DerDakon:
dashes-5-6

With "dashes: 5,6" it is better, but I think it still needs more.

DerDakon:
dashes-5-6-z15

On zoom 15 it get's worse, but I suspect this could be the renderer doing something wrong.

dashes-5-6-z16

On zoom 16 it's all fine.

rurseekatze:
I am not sure if this way of avoiding the problem is right. The current rendering of abandoned lines is good in my opinion. The only problem is that the renderer does not work correctly. So I think that we should fix the renderer to process the rendering rules correctly instead of changing the rendering. What do you think, @Dakon?

DerDakon:
Yes and no: I think the spaces should be at least the line length so it will always work even for lines with 2 tracks. But yes, the renderer needs to get debugged and fixed.

rurseekatze:
I have created an issue for debugging the renderer: https://github.com/rurseekatze/node-tileserver/issues/49.

I think that your proposed spacing change has the problem, that abandoned and razed lines are too similar, which makes it difficult for users to distinguish them. I propose to change also the rendering of razed lines to shorter dashes and more space between them.

PR #265, opened 2015-07-26
Add support for new tagging schema for Ks-Signals

PeterDRS:
Add support for new tagging schema for Ks-Signals.

### DIFF
--- a/styles/signals.mapcss
+++ b/styles/signals.mapcss
@@ -1021,9 +1021,9 @@ node|z17-[railway=signal]["railway:signal:direction"]["railway:signal:shunting"=
 	allow-overlap: true;
 }
 
-/*****************************/
-/* DE distant signal type Ks */
-/*****************************/
+/******************************************************/
+/* DE distant signal type Ks (no repeater, with Ne 2) */
+/******************************************************/
 node|z14-[railway=signal]["railway:signal:direction"]["railway:signal:distant"="DE-ESO:ks"]["railway:signal:distant:form"=light]["railway:signal:distant:states"="DE-ESO:ks1;DE-ESO:ks2"]["railway:signal:distant:shortened"="no"]["railway:signal:distant:repeated"="no"],
 node|z14-[railway=signal]["railway:signal:direction"]["railway:signal:distant"="DE-ESO:ks"]["railway:signal:distant:form"=light]["railway:signal:distant:states"="DE-ESO:ks1;DE-ESO:ks2"][!"railway:signal:distant:shortened"]["railway:signal:distant:repeated"="no"],
 node|z14-[railway=signal]["railway:signal:direction"]["railway:signal:distant"="DE-ESO:ks"]["railway:signal:distant:form"=light]["railway:signal:distant:states"="DE-ESO:ks1;DE-ESO:ks2"]["railway:signal:distant:shortened"="no"][!"railway:signal:distant:repeated"],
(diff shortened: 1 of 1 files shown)


Nakaner:
This pull request supports tags which do not conform with commonly used tags. It supports tags invented by a single user without prior discussion. I hereby propose to close this PR!

**Please discuss changes of the tagging scheme before you propose code changes.**

Nakaner:
`railway:signal:distant:additional_light=repeated/shortened` is as small-minded as our prior way to tag marker lights (Kennlicht). It is a very Germany-centric tagging and should be avoided and not supported!

PeterDRS:
How to tag it in another way?

It is a feature of the functional distant signal (no matter if it is a seperate distant signal or a "combined" signal) displaying if it is a repeator or in short distance.

What's the problem with that?

Nakaner:
We already have `"railway:signal:TYPE:shortened=yes/no` and "railway:signal:TYPE:repeated=yes/no`. There is no need for an additional tag.

PeterDRS:
As both functions are exlusive, it makes no sense to have two tags.

In this case, someone could tag both repeater and shortened which is not possible.

Nakaner:
Our tagging system is designed to be easy portable to other countries.

PeterDRS:
It is absolutely impossible - even in other countries, to be both repeater AND shortened.

Reason:
I repeater is ALWAYS shortened.

So it would be a doubled information if you tag repeater AND shortened.

BTW: The west German H/V signals don't make a difference between these two functions of the additional light. In spite of this it is not wrong to tag the concrete function of the light even at H/V signals.

So I don't really see a need to have both tags.

Nakaner:
`additional_light` is Germany-centric. Full stop.

Do you know "Halbregelabstand-Signale"? These signals can change their function from repeater to shortened and vice versa every one to two minutes.

PeterDRS:
Halbregelabstand are much more specific than Ks signals.

And in case of Halbregelabstand, the signal shows a white light, calles "Zusatzlicht". It is not specified whether it is a shortener light or repeater. But it is not a shortener in case of Halbregelabstand. Because the distant signal announces the next+1 signal.

PeterDRS:
In H/V system it would be ok to tag: railway:signal:distant:additional_light=yes which is neutral.

Nakaner:
> In H/V system it would be ok to tag: railway:signal:distant:additional_light=yes which is neutral.

With current tagging, you can easily search for signals with reduced breaking distance, e.g. using Overpass API.

PeterDRS:
If you want to search for signals with reduced braking distance it stays easy:

Just search for railway:signal:distant:additional_light.

In H/V you just HAVE NOT the information where the "Zusatzlicht" is for.

bigbug21:
PeterDRS, it is good to see you contributing to the OpenRailwayMap codebase. Your expertise in railway operations and signalling means a lot to this project. It would be good if you could lay out your proposals for tagging changes on the mailing list and I am very much looking forward to the debates to come!

PeterDRS:
This would add full support for signals having the alternate tagging scheme proposed by me.

The question is, if it is possible to activate the support just for testing/debugging purposes without answering the question which tagging variant is "right", as it seems to be a question of religion!

rurseekatze:
Your tagging proposal has not reached enough positive answers at the mailing list.

PeterDRS:
Oh my god. On the other hand it did not came any arguments against it.

bigbug21:
@PeterDRS: That is not true. There were quite some arguments given, for example in http://lists.openrailwaymap.org/archives/openrailwaymap/2015-August/000327.html. Still, you may judge those arguments in a different way. The debate came to a sudden stop and was never really finished. Speaking for myself, I apologize for not responding to the latest arguments provided.

PeterDRS:
The "debate" wasn't a debate, because everybody had the aim to show me what they think of people proposing a big change without being an "old rabbit".

Although it is technically better everybody says: "No no, never change a running system, our scheme of course is better", to not have to admit that it is not the technically better way.

So it was a fake debate to have an argument to close this issue.

PeterDRS:
BTW: The train analysis tool shows Ne2 signs as "unknown distant light signal". That's a bug, but it would have been THE CHANCE to show that tagging sign or light makes sense. It was missed. ;)

PR #266, opened 2015-07-26
Add support for new tagging scheme for Lf6/7 signals

PeterDRS:
Problem: When having two Lf7 signals at the same position showing in opposite direction, there is a hack needed having two nodes nearby.

With the proposed new tagging scheme a Lf 7 signal automatically shows in "both directions", having a nominal and reverse speed (regarding to the OSM way).

Additionally, the renderer should only render the nominal speed limit, if a reverse is tagged too.

This results in crystal clear renderings on lines with 2 tracks. Every track renders the Lf 7 only on the right track.

A solution for lines with only one track is missing here!

It is not neccessary to render Lf 6 Kennziffer 16-20, as they are not defined in Germany.

The advantage of the proposed tagging scheme railway:signal:speed_limit:announcement is to stick in only one namespace, namely railway:signal:speed_limit.

There is no neccess to have a seperate namespace railway:signal:speed_limit_distant.

Zs 3 should only be rendered when there is NO Lf 7 at the same position.


bigbug21:
It is certainly a good idea to provide a solution for the common case that both forward and backward lf7 speed signals are on the same mast. It should be discussed if "nominal" and "reseverse" in "railway:signal:speed_limit:speed:..." should rather be named "forward" and "backward". Although "nominal" and "reverse" might be better technical terms, "forward" and "backward" are very well established OSM terms.

Both issues are very well worth a debate on the mailing list.

Nakaner:
@rurseekatze Don't merge this pull request. It's too early.

rurseekatze:
I close this pull request because it contains tagging changes that were not discussed yet. Tagging changes should be proposed and discussed on the mailing list before creating pull requests.

PR #270, opened 2015-07-28
reorder signal style definitions in ascending z-index order

### LINKED ISSUE #193: Rendering of Sv signals with Sh signal
There are a lot of Sv signals in Hamburg which also have a Sh minor signal mounted below the Sv signal at the same pole. Due to a bug in the style sheet the Sh signal is rendered, not the Sv signal.

DerDakon:
The last rule that matches a given object will be used, so for objects that match multiple different rules (e.g. different types of signals at the same pole) the one with the highest z-index needs to come last.

Fixes #193.

### DIFF
--- a/styles/signals.mapcss
+++ b/styles/signals.mapcss
@@ -22,34 +22,1038 @@ meta
 	watch-modified: true;
 }
 
-/******************************************************/
-/* deactivation cross for light and semaphore signals */
-/******************************************************/
-node|z14-[railway=signal]["railway:signal:direction"]["railway:signal:main:deactivated"=yes]::deactivatedcross,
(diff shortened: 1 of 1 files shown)


Nakaner:
I think that we just have to reorder the whole style sheets (signals.mapcss and maxspeed.mapcss) if we want to fix rendering order issues like the invisible deactivation crosses and #193. We have to put more important signals at the bottom of each file. This needs almost all pull request regarding these files to be close. Otherwise conflict resolution consumes too much time. @rurseekatze

DerDakon:
According to [the the MapCSS documentation](http://wiki.openstreetmap.org/wiki/MapCSS/0.2#Order_of_Application) we need to reorder all of them, as the last rule that matches an object is taken, regardless of the z-index.

DerDakon:
The maxspeed style is not touched here, I think #282 does that better.

PR #273, opened 2015-07-29
Render Ks signals regardless of their displayable signal states

Nakaner:
This enables rendering of Ks main and combined signals which can display a marker light and implements a change of tagging of marker lights a few months ago. This fixes a bug.

### DIFF
--- a/styles/signals.mapcss
+++ b/styles/signals.mapcss
@@ -1024,14 +1024,14 @@ node|z17-[railway=signal]["railway:signal:direction"]["railway:signal:shunting"=
 /*****************************/
 /* DE distant signal type Ks */
 /*****************************/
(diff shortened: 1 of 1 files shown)


PeterDRS:
Why is ["railway:signal:distant:form"=light] a criteria for rendering?

Why are semaphore Ks signals not rendered?

DerDakon:
Because there are no semaphore Ks signals.

rurseekatze:
> Why are semaphore Ks signals not rendered?

Good joke... :smile:

PeterDRS:
Thanks for your information.

When there are no semaphore Ks signals, why tagging that it is a light signal.

And why not rendering Ks signals without the form tag?

DerDakon:
This is just a QA thing, the same as the direction tag.

PeterDRS:
It is unneccessary and should be removed.

You won't tag a post box of Deutsche Post with color:yellow.

Nakaner:
This pull request is in conflict with pull request #260. Because we should support good tagging, I close this pull request.

PR #280, opened 2015-08-01
Split preset item Bü 3/So 14, remove unnecessary comboboxes and checkboxes

Nakaner:
As written in Ril 301 of DB Netz, both Bü 3 and So 14 do not have repeaters or are placed in a shorter distance than usual. That's because they are placed *at* the activation switch.

So 14 is painted at all four faces, that's why it always gets railway:signal:direction=both.

This pull request also adds translation for these two items.


Nakaner:
The conflicts have been resolved. This PR can be merged now.

PR #282, opened 2015-08-03
Use fewer rules for the maxspeed style

DerDakon:
Using MapCSS eval(cond()) one can greatly reduce the number of lines needed for the maxspeed style.


DerDakon:
Please merge [PR 29 in node-tileserver](https://github.com/rurseekatze/node-tileserver/pull/29) first, then I can make this even shorter.

DerDakon:
I removed the icon part and put that in a different branch, as different types of changes are needed in the MapCSS converter for both commits.

PR #286, opened 2015-08-05
signals: do not show sh1 icon for signals that cannot show sh1

DerDakon:
Currently all non-dwarf sh signals are shown with state sh1, even if they do not have this state available.

### DIFF
--- a/styles/signals.mapcss
+++ b/styles/signals.mapcss
@@ -418,9 +418,11 @@ node|z17-[railway=signal]["railway:signal:direction"]["railway:signal:shunting"=
 	allow-overlap: true;
 }
 
-/****************************************/
-/* DE minor light dwarf signals type Sh */
-/****************************************/
+/*********************************************************/
+/* DE minor light dwarf signals type Sh                  */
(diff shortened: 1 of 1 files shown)


PeterDRS:
Do you mean "Zugdeckungssignale"?

Otherwise it makes no sense to have a shunting signal without state Sh 1/Ra 12. Do you have examples?

DerDakon:
Possibly. It's just that those signals that are tagged with states, but without sh1 state, should not have such an icon.

PeterDRS:
I understand how it is meant. The question for me is where those signals exist in reality.

Nakaner:
"Zugdeckungssignale" are tagged `railway:signal:minor=DE-ESO:sh` + `railway:signal:minor:states=DE-ESO:hp0;DE-ESO:kennlicht`. All of these signals I have seen up to now are no dwarf signals. But I have not seen every German Zudeckungssignal. ;-)

PeterDRS:
Of course they are not, because it is not permitted by the rules.

Nakaner:
I have checked the code in JOSM and suggest to merge this PR.

PR #289, opened 2015-08-07
use stricter regular expressions to match signal states

DerDakon:
Make sure that the given signal state is an entire state and not just a random substring in the states. Also require states for DE-ESO signals to have DE-ESO: prefix.

### DIFF
--- a/styles/signals.mapcss
+++ b/styles/signals.mapcss
@@ -1028,8 +1028,7 @@ node|z14-[railway=signal]["railway:signal:direction"]["railway:signal:distant"="
 /* DE main semaphore signals type Hp which */
 /*  - have no railway:signal:states=* tag  */
 /*******************************************/
-node|z14-[railway=signal]["railway:signal:direction"]["railway:signal:main"="DE-ESO:hp"]["railway:signal:main:form"=semaphore][!"railway:signal:main:states"],
-node|z14-[railway=signal]["railway:signal:direction"]["railway:signal:main"="DE-ESO:hp"]["railway:signal:main:form"=semaphore]["railway:signal:main:states"]["railway:signal:main:states"!~/^(.*;)?DE-ESO:hp(1|2)(;.*)?$/]
+node|z14-[railway=signal]["railway:signal:direction"]["railway:signal:main"="DE-ESO:hp"]["railway:signal:main:form"=semaphore]["railway:signal:main:states"!~/^(.*;)?DE-ESO:hp(1|2)(;.*)?$/]
 {
 	z-index: 10000;
 	text: "ref";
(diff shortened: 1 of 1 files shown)


DerDakon:
If the list [operator ~=](http://wiki.openstreetmap.org/wiki/MapCSS/0.2/Proposal_lists) is supported that one could be used instead.

styles/signals.mapcss
```diff
@@ -1306,7 +1304,7 @@ node|z14-[railway=signal]["railway:signal:direction"]["railway:signal:combined"=
 /*  - can display Hl 12a, Hl 12b           */
 /*  - cannot display Hl 11                 */
 /*******************************************/
-node|z14-[railway=signal]["railway:signal:direction"]["railway:signal:combined"="DE-ESO:hl"]["railway:signal:combined:form"=light]["railway:signal:combined:states"]["railway:signal:combined:states"!~/hl11/]["railway:signal:combined:states"=~/hl12b/]["railway:signal:combined:states"=~/hl12a/]
+node|z14-[railway=signal]["railway:signal:direction"]["railway:signal:combined"="DE-ESO:hl"]["railway:signal:combined:form"=light]["railway:signal:combined:states"]["railway:signal:combined:states"!~/^(.*;)?hl11(;.*)?$/]["railway:signal:combined:states"=~/^(.*;)?DE-ESO:hl12b(;.*)?$/]["railway:signal:combined:states"=~/^(.*;)?DE-ESO:hl12a(;.*)?$/]
```
rurseekatze: The prefix `DE-ESO` is missing.

styles/signals.mapcss
```diff
@@ -1349,8 +1347,8 @@ node|z14-[railway=signal]["railway:signal:direction"]["railway:signal:combined"=
 /*  - can display Hl 11, Hl 12a               */
 /*  - don't have to be able to display Hl 12b */
 /**********************************************/
-node|z14-[railway=signal]["railway:signal:direction"]["railway:signal:combined"="DE-ESO:hl"]["railway:signal:combined:form"=light]["railway:signal:combined:states"]["railway:signal:combined:states"=~/hl12a/]["railway:signal:combined:states"=~/hl11/]["railway:signal:combined:states"!~/hl12b/],
-node|z14-[railway=signal]["railway:signal:direction"]["railway:signal:combined"="DE-ESO:hl"]["railway:signal:combined:form"=light]["railway:signal:combined:states"]["railway:signal:combined:states"=~/hl12a/]["railway:signal:combined:states"=~/hl11/]["railway:signal:combined:states"=~/hl12b/]
+node|z14-[railway=signal]["railway:signal:direction"]["railway:signal:combined"="DE-ESO:hl"]["railway:signal:combined:form"=light]["railway:signal:combined:states"]["railway:signal:combined:states"=~/^(.*;)?DE-ESO:hl12a(;.*)?$/]["railway:signal:combined:states"=~/^(.*;)?DE-ESO:hl11(;.*)?$/]["railway:signal:combined:states"!~/^(.*;)?DE-ESO:hl12b(;.*)?$/]
```
rurseekatze: `,` is missing.

DerDakon: In fact, the following line is just wrong, as it is the opposite as this one regarding Hl12b. It's the same error as we already had for Hl3b.

PR #290, opened 2015-08-08
add a Makefile containing a simple rule to check preset syntax

DerDakon:
This will catch syntax errors like bad tag nesting.

### DIFF
--- a/josm-presets/Makefile
+++ b/josm-presets/Makefile
@@ -1,0 +1,4 @@
+.phony: check
+
+check: *.xml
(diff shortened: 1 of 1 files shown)


rurseekatze:
Note: xmllint is not yet installed on the server as it is not available in the package repos.

DerDakon:
You only need that before merging to verify that the new branch doesn't break anything ;)

rurseekatze:
Yes, I already realized that.

PR #303, opened 2015-08-17
validator: add validator for Zs3 and Zs3v states

DerDakon:
This only works in JOSM if you have at least version 8535 because of [JOSM bug 11598](https://josm.openstreetmap.de/ticket/11598).

### DIFF
--- a/validator/de-openrailwaymap.validator.mapcss
+++ b/validator/de-openrailwaymap.validator.mapcss
@@ -293,3 +293,69 @@ node[railway=signal]["railway:signal:minor"="DE-ESO:sh"]["railway:signal:minor:f
 	assertNoMatch: "node railway=signal railway:signal:minor=DE-ESO:sh railway:signal:minor:form=semaphore railway:signal:minor:states=DE-ESO:sh0";
 	assertNoMatch: "node railway=signal railway:signal:minor=DE-ESO:sh railway:signal:minor:form=semaphore railway:signal:minor:states=DE-ESO:sh0;DE-ESO:sh1";
 }
+
+/* Check correct tagging of Zs3/Zs3v signs */
+node[railway=signal]["railway:signal:speed_limit_distant"="DE-ESO:zs3v"]["railway:signal:speed_limit_distant:form"=sign]["railway:signal:speed_limit_distant:speed"]["railway:signal:speed_limit_distant:speed"!~/^[1-9][0-9]?[05]$/]["railway:signal:speed_limit_distant:speed"!=5],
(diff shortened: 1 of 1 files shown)


PeterDRS:
Zs 3 can only occur with the following values:
16, 15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2

Zs 3v can only occur with the following values:
15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2

Nakaner:
@PeterDRS http://images.deep-resonance.net/image/eisenbahn-museum-bochum-dahlhausen.jpg

PeterDRS:
It's a museum and nowadays not a valid signal.

Why should the validator support deprecated museum signals?

DerDakon:
RiL 301, 301.0301, section 5:

    (5) […] Die Kennziffern 1 und 2 können anzeigen, dass besonders früh zu halten oder in ein besetztes Gleis einzufahren ist.

And I doubt that one signal may show 16, but not the other. So all numbers from 1 to 16 (i.e. all speeds from 10 to 160) are valid, and possibly also the .5 steps in between.

rurseekatze:
The signal is located inside a museum, but it is a valid signal. As you can see on the picture above, it is still in use, so it cannot be deprecated. Museum railways are subject to the same laws as "big" railroads.

PeterDRS:
A Zs3 "16" does not need to be announced, as 160 km/h is the highest maxspeed.

It occurs only as a standalone signal.

Ril 819, module 819.0204, section 4:

> (8) Die am Signal Zs 3 bzw. Zs 3v angezeigte Geschwindigkeit
wird durch eine Kennziffer/Kennzahl (Kennziffer/
Kennzahl 2 bis 16, entspricht 20 km/h bis 160 km/h)
dargestellt.

PeterDRS:
> The signal is located inside a museum, but it is a valid signal. As you can see on the picture above, it is still in use, so it cannot be deprecated. Museum railways are subject to the same laws as "big" railroads.

Are you sure that it is a real operation under EBO/ESO in the museum? I don't think so.

bigbug21:
There are some quite odd speed signals out there. At Lutherstadt Wittenberg station, a "0 5" speed signal (announced a 5 km/h speed limit back in 2013). See http://www.openrailwaymap.org/?lang=de&lat=51.870471425392026&lon=12.664985954761503&zoom=18&style=maxspeed. I am not sure, though, if it has been a zs3.

DerDakon:
If you bring me a picture of a currently active 0.5 I'll fix the expressions ;)

PeterDRS:
@DerDakon Such vielleicht mal hier: http://www.openstreetmap.org/node/3146538234

DerDakon:
"5" as a value on signs is now allowed, and the ordering for light signals is not checked anymore.

Nakaner:
I have checked this pull request and suggest to merge it.

PR #306, opened 2015-08-19
German signal presets: reorder "off" in the list of possible states

DerDakon:
The upcoming rendering and validator rules expect off to be at the end of the list, so make the presets add them there.

### DIFF
--- a/josm-presets/de-signals-eso.xml
+++ b/josm-presets/de-signals-eso.xml
@@ -1873,9 +1873,9 @@ See http://wiki.openstreetmap.org/wiki/OpenRailwayMap for details.
 						de.text="Geschwindigkeit (km/h)"
 						delete_if_empty="true"
 						delimiter=";"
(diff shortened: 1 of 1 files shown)


DerDakon:
We discussed this, and it probably doesn't make sense to make any ordering constraints for the list of values.

PeterDRS:
An unordered list should not be wrong, but it looks nicer and is more logical to order it by speeds:
off is the highest speed, so it is logic to have it at the end of the list.
Another option would be to order it vice versa, i.e. "off;80;70;60".

There is just one exception: In H/V system a Hp2 without a Zs 3 ("off") is slower than a Hp 1 with a Zs 3 with > 60.

DerDakon:
"off" can be at either end, so this will just change an arbitrary selection. The upcoming list-operators in node-tileserver should make this work with any ordering.

PR #313, opened 2015-08-22
Ne14/Blockkennzeichen

PeterDRS:
A Blockkennzeichen is used with LZB or ETCS.

It does NOT have an Ne14.

The Ne14 is only used at virtual main signals.

A "Blechrakete" is a Zs10!

### DIFF
--- a/josm-presets/de-signals-eso.xml
+++ b/josm-presets/de-signals-eso.xml
@@ -2463,14 +2463,14 @@ See http://wiki.openstreetmap.org/wiki/OpenRailwayMap for details.
 					delete_if_empty="true" />
 			</item>
 			<item name="ETCS-Halt-Tafel (Ne 14, ETCS Stop Marker)" icon="etcs-stop-marker-arrow-left-32.png" type="node">
-				<label text="ETCS-Halt-Tafel (Ne14), auch bekannt als ETCS Stop Marker oder Blechrakete" />
-				<label text="Wird als Punkt auf dem Gleis erfasst." />
+				<label text="ETCS-Halt-Tafel (Ne14), auch bekannt als ETCS Stop Marker" />
+				<label text="Wird als Punkt auf dem Gleis erfasst. Steht am Hauptsignal oder virtuellem Signal (Level 2 ohne Signale)." />
 				<space />
 				<key key="railway"
 					value="signal" />
 				<text key="ref"
-					text="Block reference"
-					de.text="Blockbezeichner"
+					text="Signal ref"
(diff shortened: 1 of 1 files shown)


rurseekatze:
Your changes are almost OK, but why do you delete `railway:signal:train_protection:type=block_marker`? This tag should not be removed.

PeterDRS:
Because an Ne 14 is almost standing on places where it is NOT a simple block marker.

An Ne 14 stands there where a real signal would stand if the track had signals. It is NOT a replacement for the "Blockkennzeichen".

rurseekatze:
But a Ne 14 marks the position of signals (if the track would have signals) and these signals limit the block sections of the track, so they are block markers, right?

I know that Ne 14 signs are not 100% equal to "Blockkennzeichen", but the tag `railway:signal:train_protection:type=` is just intended for a rough classification of the signal. The exact meaning can be taken from `railway:signal:train_protection=`.

PeterDRS:
Ne14 signs are the opposite of block markers.

In Germany there are "Zugfolgeabschnitte", but "Blockabschnitte" are usually only called such without switches.

rurseekatze:
So if you think that `block_marker` is not a proper value for `railway:signal:train_protection:type=*`, which value would you propose? (I know that this is going to be a tagging discussion which should not be discussed here, but I am not a friend of discussions about names, so we should keep this discussion short)

PeterDRS:
railway:signal:train_protection=DE-ESO:ne14 is unique.

The Ne14 is a marker board for a virtual main signal.

I propose:
railway:signal:main=virtual

rurseekatze:
A Ne 14 is **not** a main signal. And the tag `railway:signal:main=virtual` is just nonsense.

`railway:signal:train_protection:type=virtual_signal` would be a possible solution in my opinion, but I would ask the mailing list first.

PeterDRS:
An Ne14 is a virtual **main** signal.

The tag railway:signal:main=virtual is **not** nonsense. It is the truth.

Nakaner:
@rurseekatze @PeterDRS Please move your tagging discussion to the mailing list.

@PeterDRS I will answer to you if you move this discussion to the right location (the mailing list).

PeterDRS:
@Nakaner Oh, you want to answer me. I did not even ask you.

I do not need your "knowledge" regarding railway operation.

rurseekatze:
Committed the useful changes manually in https://github.com/rurseekatze/OpenRailwayMap/commit/99fb41b30a3c0060ef06eb0b659ca675faab1be4.

PR #314, opened 2015-08-23
Use descriptions for Hl signal states

ghost:
Support users by adding description of the meaning of the state to make it possible to decide which state to tag without having a look to an external helper website or Ril 301.

Additionally, provide a multiselect for Hl distant, as they can have only one lamp (yellow) or two (yellow + green), resulting in different possible states.


Nakaner:
:+1:

HH882015:
Why is this not merged?

Nakaner:
It is invalid syntax (`display_values` and `values` have a different number of elements). I suggest to close the issue because its author has left Github.

PR #338, opened 2015-09-09
validator: suggest "Track #" to be put in description of platforms

DerDakon:
It should nowhere be put in name, but for platforms it makes some sense to put
it in description.


Nakaner:
I have checked this pull request and suggest to merge it.

PR #340, opened 2015-09-10
Add buffer stop ref

PeterDRS:
Many buffer stops have ref plates

### DIFF
--- a/josm-presets/de.xml
+++ b/josm-presets/de.xml
@@ -2435,6 +2435,11 @@ See http://wiki.openstreetmap.org/wiki/OpenRailwayMap for details.
 					values="DE-ESO:sh0,DE-ESO:sh2"
 					display_values="Sh 0-Tafel,Sh 2-Tafel"
 					default="" />
+				<text key="ref"
+					text="Reference"
+					de.text="Bezeichnung"
+					default=""
+					delete_if_empty="true" />
 				<key key="railway:signal:minor:form"
 					value="sign" />
 			</item>


rurseekatze:
The ref plates at buffer stops indicate the track ref, not a buffer ref. So I think that it is sufficient to tag the track ref on the track.

PeterDRS:
I want to tag a ref to the buffer stop according to "tag what is on the ground". Sometimes it is different, so simply documentate what's on the buffer stop. The reason is that some buffer stops are recycled and the ref plate is not changed.

rurseekatze:
Sure, we map what is on the ground. But we need to abstract the information that we see in reality. Example: We also do not map street name signs, but tag the name on the way representing the street.

PeterDRS:
Bad example, because the buffer stop is a node in OSM, the name sign not. BTW, bad example II: The city speed limit sign has a "name" tag, although the town/village is an area, too.

So 1:0 for you, of course it is fully senseless to tag a "ref" to a buffer stop. I propose to delete the ref from signals, too, because they also can have to do with the track_ref.

bigbug21:
@rurseekatze: I'm a bit confused. Sometimes we are supposed to tag what is out there, then we are up for interpretation again. I can't really see a clear course around here. If folks have fun tagging refs to buffer stops, why not? And PeterDRS is right that these signs do differ from the track refs quite often. It would be good to have this pull request merged.

rurseekatze:
But how should an on-the-ground-mapper know that the ref on the buffer stop differs from the correct track ref? In most cases he doesn't.

rurseekatze:
Nobody can prevent mappers to tag refs on buffer stops. Those mappers who want to tag that can do it manually without using a JOSM preset. But we do not have to support that by adding this tag to our JOSM presets.

bigbug21:
@rurseekatze: Looking at the sign (ref) of a buffer stop is usually easier than knowing the actual ref of a track. Therefore, and including the above-mentioned on-the-ground principle, the ref should be added to the buffer stop first, preferably using the JOSM preset.

PeterDRS:
The track ref CANNOT be taken from the ground, one  MUST look at illegal databases. That is not a joke, the track ref is the only thing that is not visible out there. Switch numbers are visible, you do not need the database here - someone here (http://forum.openstreetmap.org/viewtopic.php?id=32461&p=2) wrote wrongly:
> Noch dazu handelt es sich bei vielen Weichennummern um virtuelle Bezeichnungen, welche in der Relatiät gar nicht ersichtlich sind. Nur im Stellwerk oder auf solchen bahneingenen Plänen.

bigbug21:
@PeterDRS: You may simply know track numbers or might infer from your own general railway knowledge that the refs on display on the buffer stops are plausible track numbers.

PeterDRS:
You are very Nazi like, Mister Rurseekatze!

PeterDRS:
I will start an own project together with @bigbug21. Your "Rotz" is bullshit!

PR #344, opened 2015-09-13
detect German distant signals tagged both repeated and distant

### LINKED ISSUE #319: Throw warning: Repeated and shortened
Due to a buggy tagging scheme it is possible to tag a distant as repeated and shortened. That's of course nonsense.

It would be good to have a warning here.

DerDakon:
This should fix #319.

### DIFF
--- a/validator/de-openrailwaymap.validator.mapcss
+++ b/validator/de-openrailwaymap.validator.mapcss
@@ -225,17 +225,40 @@ node["railway:signal:train_protection"="DE-ESO:ne14"]["railway:signal:train_prot
 
 /* a German distant signal can't be a repeated one if at the same pole as a main signal, as the prior distant
  * signal would refer to this main signal */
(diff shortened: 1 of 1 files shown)


Nakaner:
You know that we already have this rule?

    node["railway:signal:main"="DE-ESO:hp"]["railway:signal:distant:repeated"="yes"]
         ["railway:signal:main:states"!~/^(.*;)?DE-ESO:kennlicht(;.*)?$/]

If I have a signal tagged with `railway:signal:main=DE-ESO:hp`, `railway:signal:distant=DE-ESO:vr`, `railway:signal:main:states=DE-ESO:hp0;DE-ESO:hp1`, `railway:signal:distant:repeated=yes`, `railway:signal:distant:shortened=yes`, I get the error message mentioned above in addition to your error message.
> main and repeated distant signal usually are not at the same place

DerDakon:
I have adapted the other rule, too, to avoid both matching at the same time.

Nakaner:
I suggest to merge this pull request. I have tested it.

Issue #276, opened 2015-07-31
Render *.0 milestones


ghost:
The milestones with km *.0 should be handled special, in form that they are rendered even in lower zoom levels and get a better z-index.

That gives a better overview over the track, when needing finer stations, zoom in and then seeing all milestones.

Maybe to save space the label could be cropped from "76.0" to "76".

HH882015:
Close this

rurseekatze:
I do not like the idea to crop e.g. `76.0` to `76` because that could it make more difficult for users to identify that label as a kilometer value.

PR #345, opened 2015-09-14
style: use higher z-index for milestones on full kilometers

DerDakon:
In case all milestones are tagged those should give a better rendering as the full kilometers are more likely to appear.

This implements half of #276.


rurseekatze:
In most cases, these changes do not have a visible effect because many milestones are not rendered due to collisions with more important `ref=*` labels.

Issue #321, opened 2015-08-27
Validation rules for bridges and tunnels


rurseekatze:
At bridges or tunnels, some tags should have a prefix so that the information is related to the bridge/tunnel instead of the track:

* way with `tunnel=*` and `name=*` containing "tunnel" -> propose `tunnel:name=*`
* way with `bridge=*` and `name=*` containing "bridge", "brücke", "viadukt", etc. -> propose `bridge:name=*`
* way with `tunnel=*` and `wikipedia=*` and not `tunnel:wikipedia=*` -> propose `tunnel:wikipedia=*`
* way with `bridge=*` and `wikipedia=*` and not `bridge:wikipedia=*` -> propose `bridge:wikipedia=*`

PR #346, opened 2015-09-15
validator: suggest changing name to bridge:name or tunnel:name if it contains certain keywords

DerDakon:
This implements parts of #321.


Nakaner:
I have tested it. It works. (JOSM rev. 8800)

PR #351, opened 2015-09-17
reorder tag checks for signal taggings

DerDakon:
If the same keys are checked in multiple ways move them to the end of the line, so the constant parts are together in all lines.


DerDakon:
Please merge #361 first.

PR #353, opened 2015-09-23
Update de-signals-eso.xml

sb12:
forgotten tag when moving from `lzb_start` to `train_protection`


sb12:
There's another one here: https://github.com/rurseekatze/OpenRailwayMap/blob/master/styles/josm-additional.mapcss#L78 but I'm not sure what would be the correct expression.

Nakaner:
:+1: 

@sb12 wrote:
> There's another one here: https://github.com/rurseekatze/OpenRailwayMap/blob/master/styles/josm-additional.mapcss#L78 but I'm not sure what would be the correct expression. 

This rule renders the deprecation icon (`misc/deprecated.png`) from JOSM's own iconset if an object has a deprecated tag, i.e. it's right.

PR #355, opened 2015-09-24
remove parts of obsolete rendering code

DerDakon:
These files are either identical to those in renderer (i.e. node-tileserver) or
are in a broken state while the file in renderer works. Fix the links and
remove the old files altogether.


rurseekatze:
Can you also remove `js/kothic.js`, `js/kothic-leaflet.js`, `js/style` and `js/utils` and fix the necessary paths (e.g. in `canvas.php`)?

DerDakon:
Those are actually different than what is on server (compared to the version in node-tileserver):
 - kothic.js
 - style/mapcss.js
 - utils/collisions.js

DerDakon:
Done.

PR #357, opened 2015-09-25
improve states multiselect for H/V signals

sb12:
This solves some problems with the multiselect in the josm preset for the H/V signal states

- JOSM currently orders multiselects alphabetical according to the name. This means the ordering is different for each language. So if somebody edits a Hp signal with a German JOSM and the next one edits it with an English JOSM the values are swapped eventhough nothing changed. I changed the names, so that they are in a logical order regardless of the language. (I didn't find another possibility to sort multiselects, if somebody knows a better way please let me know)

- The value `off` (dunkel geschaltet) was missing. For Vr signal this value occurs when there is no corresponding Hp signal on the route. I think it can also occur for Hp and Vr signal on high speed lines with LZB installed, though I haven't seen this in reality for H/V signals, yet.

- The value `?` was also missing. This is used when it is e.g. not known whether a signal can show Hp 1 and Hp 2 or just one of them.


Nakaner:
I have seen Hp and Ks main signals being dark when passing them with a LZB-directed train.

DerDakon:
I don't like the "ESO" prefix for Vr signals, although I understand why you added them.

sb12:
Best would be to solve the sorting issue in josm. I have created/reopened the issues https://josm.openstreetmap.de/ticket/11921 and https://josm.openstreetmap.de/ticket/5509

sb12:
> I don't like the "ESO" prefix for Vr signals, although I understand why you added them.

I'm not very happy with them either. Does anybody have an idea for a better method to have the vr in front of the kennlicht etc.?

If the ESO is a reason for not merging it, I can also remove it, which then leads to the vr showing up last, behind the ?, marker light and off.

Klumbumbus:
JOSM has now the option to disable alphabetic sorting for a single key. see https://josm.openstreetmap.de/ticket/11926

sb12:
I will adapt the pull request to use the new option as soon as I find some time to install the latest JOSM version. Please wait with merging until then.

sb12:
Updated with the new JOSM option and ready to merge now. Works in the latest stable JOSM version.

Thanks to the JOSM team for the quick solution.

PR #358, opened 2015-09-26
Render tracks tagged with maxspeed given in mph

### LINKED ISSUE #84: Render maxspeeds stated in mph
At the moment, only speeds in km/h are rendered. Is it possible to also render speeds stated in mph? The UK is looking rather dull at the moment...

DerDakon:
Needs some patches in node-tileserver, but with them I'm able to see a proper rendered map. Only those lines with a maxspeed as a multiple of 5 mph are colored.

Merge #282 first. Fixes #84.


bigbug21:
Thanks!

PR #368, opened 2015-10-04
signal style: show train protection systems of tracks

DerDakon:


DerDakon:
leipzig-train-protection

Issue #67, opened 2014-10-17
Continously provide stable links in URL line while browsing the map


bigbug21:
In order to make sharing OpenRailwayMap links more conveniant (and thus more popular) the URL to the current map view should be continously updated. Just like the main OpenStreetMap map one could easiler copy&paste the URL for the current map view from the browser's URL line.

bigbug21:
Two thumbs up and good luck for OSM Sommercamp!

hiddewie:
This feature has been implemented in the OpenRailwayMap vector fork, see e.g. https://openrailwaymap.fly.dev/#view=10.99/48.8658/2.3383&style=signals and move the map around.

PR #374, opened 2015-10-14
show proposed and constructed tracks only when the finished tracks would be shown

DerDakon:
Otherwise e.g. proposed tram lines would be shown on level 9, but tram lines only on level 11.


sb12:
Same problem applies to abandoned, disused and razed tram lines. Maybe you could apply the same change to these as well.

Good test location is [Karlsruhe](http://www.openrailwaymap.org/?lang=en&lat=49.009501205463394&lon=8.581008911132812&zoom=10&style=standard) where you have abandoned, disused, razed, proposed and constructed tram lines.

DerDakon:
Working on it. This needs to be done for bridges and tunnels, too, so this becomes a bit messy.

DerDakon:
This will only be fixed if [#67 in node-tileserver](https://github.com/rurseekatze/node-tileserver/pull/67) will be merged first as it get's unnecessarily messy otherwise.

DerDakon:
This need #399 to work.

PR #388, opened 2015-10-19
signal layer: the property name is font-size, not text-size

DerDakon:
It looks to me the default font size used when font-size not explicitely set is 10, so this may actually lead to different rendering. I can change this to 10 or just remove those lines whereever you want.

### DIFF
--- a/styles/signals.mapcss
+++ b/styles/signals.mapcss
@@ -296,7 +296,6 @@ node|z14-[railway=signal]["railway:signal:direction"]["railway:signal:main"="AT-
 	z-index: 1000;
 	text: "ref";
 	text-offset: 11;
-	text-size: 12;
 	text-color: black;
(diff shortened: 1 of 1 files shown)


Nakaner:
With your changes, the font in JOSM becomes about 1.5 times larger, i.e. you should change the from 12 to 10.

DerDakon:
The question is if it's not better to just remove those lines as they were actually never used.

Nakaner:
@DerDakon  wrote:
> The question is if it's not better to just remove those lines as they were actually never used.

I agree with you. The current font size is correct. Larger labels only need more space on the map.

DerDakon:
Ok, since the value had no effect and the test is still rendered (checked e.g. on Blockkennzeichen) I just removed it.

Issue #61, opened 2014-10-13
Search engine optimization


rurseekatze:
The HTML code should be optimized, so that the website can be found with search engines.

It is also necessary to improve the texts because the description/extract that is currently shown by e.g. Google is not very informative.

Issue #60, opened 2014-10-12
Add OpenRailwayMap do Taginfo projects list


endro:
http://wiki.openstreetmap.org/wiki/Taginfo/Projects

rurseekatze:
Was fixed in https://github.com/rurseekatze/OpenRailwayMap/commit/c68756d67af7c871a5638ec03bfd8cb924ee7fa1.

This file is far away from being complete. This will need some time and work. Contributions are welcome.

The integration into https://github.com/taginfo/taginfo-projects will be done soon.

PR #398, opened 2015-10-25
signal style: use @media instead of kothicjs-ignore-offset-x

DerDakon:
This has the same effect without a custom pseudo property.

While at it merge some code for DE-ESO:blockkennzeichen at different zoom levels.

### DIFF
--- a/styles/signals.mapcss
+++ b/styles/signals.mapcss
@@ -319,8 +319,7 @@ node|z17-[railway=signal]["railway:signal:direction"]["railway:signal:stop"="DE-
 	text-halo-color: white;
 	/* text-offset-x: eval(add(mul(length(tag("railway:signal:stop:caption")), 3), 6)); */
 	text-offset-x: -12;
-	kothicjs-ignore-text-offset-x: true;
-	text-offset-y: 12;
+	text-offset-y: -12;
 	icon-image: "icons/de/ne5-ds301-32.png";
 	icon-width: 11;
 	icon-height: 16;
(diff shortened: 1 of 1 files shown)


DerDakon:
This needs [#60](https://github.com/rurseekatze/node-tileserver/pull/60) and [#61 in node-tileserver](https://github.com/rurseekatze/node-tileserver/pull/61) to work.

styles/signals.mapcss
```diff
@@ -730,13 +737,11 @@ node|z16-[railway=signal]["railway:signal:direction"]["railway:signal:train_prot
...
+node|z16-[railway=signal]["railway:signal:direction"]["railway:signal:train_protection"="DE-ESO:blockkennzeichen"]::blockkennzeichen
 {
 	z-index: 8510;
 	text: "ref";
 	text-offset-x: -28;
-	kothicjs-ignore-text-offset-x: true;
-	text-offset-y: 12;
 	text-color: black;
 	text-halo-radius: 1;
 	text-halo-color: white;
```
rurseekatze: You could remove `text-halo-color` and `text-halo-radius` as the background of the caption is a white sign.

styles/signals.mapcss
```diff
@@ -730,13 +737,11 @@ node|z16-[railway=signal]["railway:signal:direction"]["railway:signal:train_prot
 	icon-height: 20;
 	allow-overlap: true;
 }
-node|z16-17[railway=signal]["railway:signal:direction"]["railway:signal:train_protection"="DE-ESO:blockkennzeichen"]::blockkennzeichen
+node|z16-[railway=signal]["railway:signal:direction"]["railway:signal:train_protection"="DE-ESO:blockkennzeichen"]::blockkennzeichen
 {
 	z-index: 8510;
 	text: "ref";
 	text-offset-x: -28;
```
rurseekatze: `text-offset-x` should be set to `-10` to align the text in the middle of the sign in JOSM.

styles/signals.mapcss
DerDakon: If the text fits inside the icon, yes. But it will look awkward otherwise anyway.

Issue #387, opened 2015-10-18
Missing icons of JOSM preset


Klumbumbus:
The JOSM presets use icons, which are not availible, which produces errors in the JOSM statusreport, see https://josm.openstreetmap.de/ticket/11980#comment:6

don-vip:
Hi,
We have now a unit test that detects invalid presets:
https://josm.openstreetmap.de/jenkins/job/JOSM/jdk=JDK7/lastUnsuccessfulBuild/testReport/junit/org.openstreetmap.josm.gui.preferences.map/TaggingPresetPreferenceTest/testValidityOfAvailablePresets/

OpenRailwayMap ones generate a lot of errors due to missing icons.

Can you please provide a zip archive containing all the missing icons?

don-vip:
The Makefile should also detect missing icons, like in f7783567ed1ef2124b354319a9f9f1cbbf2b56ab for styles.

rurseekatze:
In https://github.com/rurseekatze/OpenRailwayMap/commit/f4778127b4402f4f97406952ae92daf9f9a33835 I have extended the existing Makefile, so that now it can generate zip archives containing all icons.

The new zip archives are available under http://www.openrailwaymap.org/josm-presets/. I have already updated the URLs of our presets  at https://josm.openstreetmap.de/wiki/PresetsSource.

Currently the Makefile simply adds all icons to the archive, ignoring whether they are used or not. In a later version, it will parse a list of all used icons from the preset file and add only these icons to the archive for generating smaller files.

> The Makefile should also detect missing icons, like in f778356 for styles.

If the unit test detects no problems any more, I would close this issue and create a new one for extending the Makefile to detect missing icons.

don-vip:
Thanks! :)

don-vip:
There are still some errors.

DerDakon:
Some icon references were fixed in #407. What's definitely missing are icons for the Sk signals, they just do not exist. In #409 I have changed the generation to only include those icons in the ZIPs that are actually referenced.

rurseekatze:
There were still a lot of items with invalid icon references. The items had placeholder icon references too, but they were already the names of the planned icons instead of `todo.png`.

In https://github.com/rurseekatze/OpenRailwayMap/commit/550a77ac3f9421a10a914ef19d7b464bb5ff7430 I removed all of these icon references. Now there should not be any errors anymore.

rurseekatze:
Can this issue be closed?

[...]

PR #405, opened 2015-11-04
clean up JOSM templates and improve verification

DerDakon:
This drops some unneeded keys from the JOSM presets. Those settings are the default since [JOSM version 5155](https://josm.openstreetmap.de/changeset/5155/josm) anyway. At one place one of those tags was even at an illegal position, resulting in the AVG presets not being loadable without error. Finally the check is improved to use the JOSM XML schema to better verify the templates.

### DIFF
--- a/josm-presets/Makefile
+++ b/josm-presets/Makefile
@@ -1,4 +1,4 @@
 .phony: check
 
 check: *.xml
-	@xmllint --noout *.xml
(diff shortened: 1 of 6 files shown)


don-vip:
This is a great change, thank you! Now #387 is the only remaining error we detect in JOSM unit tests.

PR #412, opened 2015-11-12
JOSM presets: drop presets that are already offered by JOSM itself

DerDakon:
man_made=works may even be incorrect, and since it is not railway specific in any way drop it.


rurseekatze:
Can you update the Austrian preset file too?

DerDakon:
Done.

PR #414, opened 2015-11-14
Remove Defaults which May Impair Data Quality

Nakaner:
Example: You often cannot recognize if a distant signal is a repeater or not while passing it with a fast train. That's why this checkbox should not have a default value.


rurseekatze:
The problem also occurred in other preset files: https://github.com/rurseekatze/OpenRailwayMap/commit/18778bda612d1581b30b38719eeab7236f9e149e.

PR #420, opened 2015-11-21
set 'font-weight: bold' where a bold font-family is requested

DerDakon:
This currently is automatically added by KothicJS font processing, but better be explicit about what is requested.

### DIFF
--- a/styles/standard.mapcss
+++ b/styles/standard.mapcss
@@ -62,6 +62,7 @@ way|z12-[bridge=yes][!"railway:track_ref"].tracks::bridges
 	text-color: #585858;
 	font-size: 11;
 	font-family: "Verdana Bold";
+	font-weight: bold;
 	font-style: italic;
 	text-halo-radius: 1;
 	text-halo-color: white;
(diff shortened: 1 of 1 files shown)


rurseekatze:
Strangely this change has the effect that tunnel and bridge labels are not rendered any more if the tiles are rendered on server side. This error does not occur for client side rendering.

DerDakon:
Not that I know what could cause this, but is "Verdana Bold" actually installed on the server? Otherwise one of the automatically added fallback fonts is used.

rurseekatze:
OK, the problem seems to be not related to the changes in this pull request. The rendering issue also occurs with the previous version of this style. So it could be caused by missing fonts on the server. I will check that.

rurseekatze:
The problem seems to be caused by `eval()` or `join()`. I removed those functions from the stylesheets and replaced them by simple strings and the labels of bridges or tunnels were rendered. I guess that something from https://github.com/rurseekatze/node-tileserver/pull/42 does not work.

I do not know why it worked on client side. Maybe an old version of the code was in the cache or the browser JS engine has a different behaviour than Node JS.

Can you help to debug it?

DerDakon:
It's the redefinition of e_join that causes trouble. Will open a PR for it in a minute.

rurseekatze:
> It's the redefinition of e_join that causes trouble. Will open a PR for it in a minute.

Was fixed in https://github.com/rurseekatze/node-tileserver/pull/77.

PR #432, opened 2015-12-05
Add .gitignors for josm-preset/ and styles/

Nakaner:


josm-presets/.gitignore
```diff
@@ -0,0 +1 @@
+*.zip
```
DerDakon: Those files are already ignored.

styles/.gitignore
```diff
@@ -0,0 +1,2 @@
+*.zip
```
DerDakon: These can be added to /.gitignore

styles/.gitignore
```diff
@@ -0,0 +1,2 @@
...
+*.js
```
DerDakon: Those files are already ignored.

PR #443, opened 2015-12-12
use working blog URL

DerDakon:


rurseekatze:
This change is too early now.

PR #445, opened 2015-12-19
use automatic loading of Flattr button

DerDakon:
This avoids the delay until the map itself is loaded I can sometimes observe.

### DIFF
--- a/canvas.php
+++ b/canvas.php
@@ -157,6 +157,7 @@
 			(function(){
 				var s = document.createElement('script'), t = document.getElementsByTagName('script')[0];
 				s.type = 'text/javascript';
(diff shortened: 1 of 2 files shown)


canvas.php
```diff
@@ -157,6 +157,7 @@
 			(function(){
 				var s = document.createElement('script'), t = document.getElementsByTagName('script')[0];
 				s.type = 'text/javascript';
+				s.mode = auto;
```
rurseekatze: This does not work. I guess you mean `s.mode = 'auto';`.

DerDakon: I'm sure I tested this and it was the other way round…

PR #448, opened 2015-12-31
style: add color to proposed and constructed railways

DerDakon:
The same color as for the operational lines is used. The dashes for railways under construction are shortened a bit to make tracks with multiple lines not appear as a solid line when the 2 tracks are "badly" aligned, e.g. by curves.


rurseekatze:
Looks good. Could you also add this for branch and main lines, please?

DerDakon:
I guess highspeed as well?

DerDakon:
Fixed.

PR #451, opened 2016-01-07
style: replace deprecated "@media" by "@supports"

don-vip:
See https://josm.openstreetmap.de/wiki/Help/Styles/MapCSSImplementation#Mediaqueriessince6970deprecated

### DIFF
--- a/styles/signals.mapcss
+++ b/styles/signals.mapcss
@@ -3,7 +3,7 @@ OpenRailwayMap Style File for JOSM and KothicJS
 Signalling layer
 
 OpenRailwayMap Copyright (C) 2012 Alexander Matheisen
-This program comes with ABSOLUTELY NO WARRANTY.
+This program comes with ABSOLUTELY NO WARRANTY.jo
 This is free software, and you are welcome to redistribute it under certain conditions.
 See http://wiki.openstreetmap.org/wiki/OpenRailwayMap for details.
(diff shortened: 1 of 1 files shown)


DerDakon:
This will not work until the mapcss parser in node-tileserver learns to support @supports.

DerDakon:
As suggested by your own documentation you should also add "min-josm-version: "8087"; /* This style uses @supports rules */" to the header of the file.

don-vip:
correct.

stoecker:
Any progress? It's the last issue causing a test failure.

DerDakon:
The code is in [my @supports-branch](https://github.com/DerDakon/node-tileserver/tree/%40supports), but not finished yet.

DerDakon:
I would suggest 2 things:
 * squash those 2 commits together, because they logically belong together
 * invert the logic, so that the JOSM code lives inside the "@supports", the current form was only used because of the constrained "@media" implementation in node-tileserver.

don-vip:
I'm not much a git guy. I tried to squeeze my commits and apparently messed up my branch. Can you please just close my PR and commmit what you think is the best? It will be faster than wait for me to understand how to git rebase properly.

don-vip:
I'm unable to merge my commits, I have created #459, easier.

Issue #56, opened 2014-09-10
Render maxspeed:forward and maxspeed:backward


rurseekatze:
Please render maxspeed:forward and maxspeed:backward in maxspeed layer.

Possible implementations:

**Equal altering of way colour**
Altering between the two colors, and speed could be written with an arrow to show the direction, like this: < 80  - 100 >

**Altering of way colour depending on direction**
A stroke of 2/3 length with the colour of maxspeed:forward, a stroke of 1/3 length with the colour of maxspeed:backward.

**Arrow icons**
Arrows: <---> <---> <--->
Colour: aaabb aaabb aaabb
a = `maxspeed:forward=*`
b = `maxspeed:backward=*`
left = forward direction of the OSM way

**Render way borders**
This solution needs the `offset` property of MapCSS.
Render the right half of the way with the color of `maxspeed:forward=*`, the left half of the way with `maxspeed:backward=*`.

PR #458, opened 2016-02-09
render maxspeed if it is only given for one direction

DerDakon:
This renders lines where maxspeed is given as maxspeed:forward or maxspeed:backward, preferring the direction given in railway:preferred_direction. The other direction is not rendered.

This may be seen as a fix for #56.

### DIFF
--- a/locales/de_DE/LC_MESSAGES/messages.po
+++ b/locales/de_DE/LC_MESSAGES/messages.po
@@ -490,3 +490,9 @@ msgstr "Mapnik ohne Beschriftungen Graustufen"
 
 msgid "Blog"
 msgstr "Blog"
+
+msgid "No information"
+msgstr "Keine Angaben"
+
+msgid "speed only given for one direction"
+msgstr "Geschwindigkeit nur für eine Richtung angegeben"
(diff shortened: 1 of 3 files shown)


rurseekatze:
In http://www.openrailwaymap.org/?lat=48.05156978288746&lon=7.727956473827361&zoom=18&style=maxspeed it does not work properly. The color of track 1 indicates 40 km/h, but the caption says 80 km/h.

Also the legend caption "speed only given for one direction" can be misunderstood. I the example, the track has both `maxspeed:forward=*` and `maxspeed:backward=*`, so it is contradictory to the legend caption. But on the other hand, the legend caption can also mean "the maxspeed rendered on the map just gives information for one direction". What was your intended meaning of the legend caption?

DerDakon:
Yes. I'm looking into the first issue.

DerDakon:
Should be fixed by #506.

rurseekatze:
> Should be fixed by #506.

:+1: 

What we still need to discuss is the second problem.

DerDakon:
"speed only shown for one direction"?

DerDakon:
"speed only relevant for one direction"?

PR #466, opened 2016-02-28
Add Finish Main and Distant Signals of Eo/Po Type

Nakaner:
These icons have been retrieved from https://commons.wikimedia.org/wiki/Category:Diagrams_of_railway_signals_in_Finland and are licensed as public domain/CC-0.


styles/signals.mapcss
```diff
@@ -1480,6 +1549,78 @@ node|z14-[railway=signal]["railway:signal:direction"]["railway:signal:main"="AT-
 	allow-overlap: true;
 }
 
+/*************************/
+/* FI main light signals */
+/* general rules         */
+/*************************/
+node|z14-[railway=signal]["railway:signal:direction"]["railway:signal:main"=~"^FI:Po(-v)?$"]["railway:signal:main:form"=light]
```
DerDakon: This regex is invalid (as well at the other location), it must be /^...$/.

PR #468, opened 2016-03-02
Show more tracks on the electricity layer

DerDakon:
Highlight deelectrified tracks and those where the electrification is only proposed or under construction.


DerDakon:
Looks like I also need to update the legend, too.

PR #480, opened 2016-03-18
Support railway=border

Nakaner:
`railway=border` tag has been invented over a year ago but has not been implemented yet.


styles/standard.mapcss
```diff
@@ -816,10 +816,11 @@ node|z14-[railway=tram_stop]
 /*****************/
 /* owner changes */
 /*****************/
-node|z11-[railway=owner_change]
+node|z11-[railway=border],
```
rurseekatze: Why do you use one icon for both border and owner change? You created different icons, so they should be used.

josm-presets/de.xml
```diff
@@ -1740,16 +1740,52 @@ See http://wiki.openstreetmap.org/wiki/OpenRailwayMap for details.
...
+				text="- railway:name:(abbrevation of operator 1) = (abbrevation of this location)"
+				de.text="- railway:ref:(Kürzel Betreiber 1) = (Abkürzung)" />
+			<label
+				text="- railway:name:(abbrevation of operator 2) = (abbrevation of this location)"
+				de.text="- railway:ref:(Kürzel Betreiber 2) = (Abkürzung)" />
+		</item>
+		<item name="Border" de.name="Grenzpunkt" icon="presets/douane.png" type="node">
+			<!-- presets/douane.png is included in JOSM's internal preset set -->
+			<label
+				text="Change of owner/operator near a boundary (admin_level>=2)"
```
rurseekatze: Shouldn't it be `admin_level<=2` instead of `admin_level>=2`?

styles/standard.mapcss
Nakaner: I wrote at [the commit message](https://github.com/rurseekatze/OpenRailwayMap/pull/480/commits/dbf2246c02999f60b158641bb1da646cd00762b2):
> It is intention to use the same icon for both owner changes and borders.
The grey-black icon would be bad visible on the map.

josm-presets/de.xml
```diff
@@ -1740,16 +1740,52 @@ See http://wiki.openstreetmap.org/wiki/OpenRailwayMap for details.
 				/>
 		</item>
 		<separator />
-		<item name="Grenzpunkt" icon="owner-change.png" type="node">
-			<label text="Eigentumswechsel der Infrastruktur an Privatbahnanschlüssen und Landesgrenzen" />
+		<item name="Change of Operator" de.name="Betreiberwechselpunkt" icon="owner-change.svg" type="node">
```
rurseekatze: The term "Betreiberwechselpunkt" sounds a bit strange, I have never heard of this. I would propose the terms "Betreiberwechsel", "Betreibergrenze" or "Anschlussgrenze" instead.

josm-presets/de.xml
```diff
@@ -1740,16 +1740,52 @@ See http://wiki.openstreetmap.org/wiki/OpenRailwayMap for details.
...
+			<!-- presets/douane.png is included in JOSM's internal preset set -->
+			<label
+				text="Change of owner/operator near a boundary (admin_level>=2)"
+				de.text="Eigentumswechsel der Infrastruktur nahe der Staatsgrenze" />
+			<key key="railway"
+				value="border" />
+			<space />
+			<label
+				text="You have to add manually following tags:"
+				de.text="Manuell sind noch folgende Tags zu ergänzen:" />
```
rurseekatze: Could you use the syntax convention we already use, please? The first attribute should be in the same line with the label element.

styles/standard.mapcss
rurseekatze: Sorry, I did not read that. It is not the best solution to use the same icon for multiple features, but at the moment it is ok as a quick solution. We can think of an alternative icon in the future.

josm-presets/de.xml
Nakaner: fixed

Nakaner:
I have fixed all the issues @rurseekatze complained about.

PR #490, opened 2016-03-29
Change Path to Level Crossing Icon

### LINKED ISSUE #489: josm preset uses icon, which is no longer availible
in line:

https://github.com/rurseekatze/OpenRailwayMap/blob/2b8004a5c39d7404065807ede94b4366fa5bca1e/josm-presets/de.xml#L1400

icon was removed in josm. new icon path is:

styles/standard/transport/railway/level_crossing.svg

Nakaner:
This fixes #489.

@rurseekatze Some saltires are mounted in Germany in this way (i.e. rotated by 90°).

### DIFF
--- a/josm-presets/de.xml
+++ b/josm-presets/de.xml
@@ -1397,7 +1397,7 @@ See http://wiki.openstreetmap.org/wiki/OpenRailwayMap for details.
 			</roles>
 		</item>
 		<separator />
-		<item name="Level Crossing" de.name="Bahnübergang" icon="presets/level_crossing.png" type="node">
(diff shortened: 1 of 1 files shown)


Klumbumbus:
In JOSM we choose the icon in "landscape orientation" becasue it seems to be this way in most countries according to https://commons.wikimedia.org/wiki/Category:Diagrams_of_warning_crosses and https://en.wikipedia.org/wiki/Level_crossing

Nakaner:
There are also some saltires in Germany There are also some saltires in Germany which are rotated by 90°. There are more important things to do than rotating an icon.

Klumbumbus:
The main reason was the change from png to svg.

PR #496, opened 2016-04-17
Refactor definitions which tracks are shown

DerDakon:
- put it in a common file
- do not show e.g. sidings unter construction on zoom level 9 while active sidings are only shown on zoom level 10

### DIFF
--- a/styles/Makefile
+++ b/styles/Makefile
@@ -11,9 +11,15 @@ all: $(JSFILES) $(ZIPFILES)
 %.js: %.mapcss
 	$(PYTHON) ../renderer/mapcss_converter.py --mapcss $^ --icons-path .
 
-%.zip: %.mapcss
-	zip $@ $^ | sed '/^[ \t]*adding: /d'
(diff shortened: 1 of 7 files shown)


Nakaner:
JOSM [does not support](https://josm.openstreetmap.de/wiki/Help/Styles/MapCSSImplementation#Grammar) `@import`. We should therefore write workaround in our build process which produces the ZIP files. This workaround shold insert the content of common.mapcss at the specified line into the MapCSS file.

Nakaner:
Nakaner wrote:
> JOSM does not support @import. We should therefore write workaround in our build process which produces the ZIP files. This workaround shold insert the content of common.mapcss at the specified line into the MapCSS file.

Could you, @DerDakon, modify the Makefile at this pull request in a way that it produces ZIP files which are compatible with JOSM? You could use following awk script (GNU awk necessary):

    #!/usr/bin/env gawk
    {
            if ($0 ~ /@import/) {importfile=gensub(/@import url\("([^"]+)"\);/, "\\1", "g")
            while ((getline line < importfile) > 0)
                    print line
            close(importfile)
            } else
                    print
    }

The script replaces the import by the content of the file to be imported.

DerDakon:
Ignore this for the moment, it has to be rebased to include the new changes in signalling and electrification layers.

Issue #106, opened 2015-01-18
Validator Rules: railway:signal*=* Without railway=signal


Nakaner:
Quote from Dakon from #OpenRailwayMap IRC:
> Reported by ich fände es toll wenn JOSM die 2 Fehler, die ich in der letzten Woche gemacht habe, 
> anmeckern würde:
>   -railway:signal* ohne railway=signal
>   -railway=signal nicht auf den Schienen, sondern daneben

In English:
* JOSM should warn if a node has railway:signal:*=* but no railway=signal.
* JOSM should warn if a node has railway=signal but is not part of a track.

rurseekatze:
A rule for detecting the first problem (railway:signal:= but no railway=signal) was added with #119.

It seems that the second problem cannot be detected yet, as the MapCSS validation rules can currently just check tagging problems, but not geometry problems.

rurseekatze:
Fixed by #500.

PR #500, opened 2016-04-21
validator: warn about signals that are not tagged on the track

DerDakon:
Fixes the remaining part of #106.


rurseekatze:
I think that `throwError` is better for this type of wrong mapping than `throwWarning`.

PR #501, opened 2016-04-27
update Leaflet to 0.7.7

DerDakon:
Drop the local copy, just use the Leaflet CDN to retrieve the code. This update removes some compat code for IE8 and older, which we do not support anymore.

I have tested this for several weeks and have not seen any ill effects on the canvas layer.


Nakaner:
It think that we should keep our local copy for privacy reasons. Every user who visits openrailwaymap.org will request Leaflet from Leaflet's CDN and therefore the CDN operator gets to know who visited ORM when (and he can do some user tracking).

If all websites use the same CDN for one and the same Javascript code snippets, all these websites will be affected if someone breaks in at the CDN and changes some code lines at Leaflet.

I know that caching does not work if every website provides its own copy of Leaflet. But once a user visits a website twice, he gets the Leaflet version from hi local cache. (Do browsers MD5sum the files in their cache and cache across domains?)

DerDakon:
Sure, but then we first should get rid of Paypal and Flattr buttons.

DerDakon:
We actually may have no other choice. There is [subresource integrity](https://www.w3.org/TR/SRI/), but that would require a [CORS](https://www.w3.org/TR/cors/) header to be present on the CDN, which is not set there.

rurseekatze:
That is a difficult decision. Both solutions have their advantages:

Using CDN:
* We can clean our repository from foreign code
* Updating Leaflet is easier (just changing the URL)
* It is more probable that the user already has a copy of Leaflet in his cache

Own hosting:
* Better privacy
* Independency from other providers

rurseekatze:
Some comments on the mentioned arguments:

> We can clean our repository from foreign code

That is not really an advantage of using a CDN because we could use the solution that was introduced by @DerDakon in this pull request. The libs are not contained in the repository and just downloaded and installed while the installation.

> Updating Leaflet is easier (just changing the URL)

If we use this solution of dropping the code from the repository and downloading it while installing, it would also just require to change the URL in the INSTALL.md or a shell script.

> It is more probable that the user already has a copy of Leaflet in his cache

This would only work if other websites are using exactly the same version of Leaflet. I want to link to a special version, not to the latest version of Leaflet to avoid that something is broken with a new version of the library. It is not very probable that other websites are using the same version of Leaflet and were visited by a user before our site. So in most cases there would not be any speed up for the user by using a local cached version. Maybe it is a bit faster in these cases because the content is loaded from different domains (browsers are limited in connections per domain, that is the reason for using subdomains like a.tiles.... for tile servers).

An advantage for us is that the server traffic and the load is reduced.

rurseekatze:
I decided to keep the local copy of Leaflet.

PR #505, opened 2016-04-30
add german BOStrab signal G4 60

sb12:
This adds the BOStrab maxspeed signal G4 60, which has been missing until now. You can find a few of them e.g. in Karlsruhe
Code is untested.


sb12:
Any idea why the "6" is bigger than all the other speed signals?
screenshot from 2016-05-15 00 16 06

DerDakon:
Because the style of the icons is defined a bit above in a separate block so it is not repeated for all icons. Fixed in #514.

sb12:
Thanks @DerDakon for fixing. I must have overlooked this block.

PR #527, opened 2016-06-30
treat railway=construction;construction=* the same as construction:railway=*

DerDakon:
This is the default tagging suggested in the wiki. Do the same change for railway=proposed. Also make sure not to render such platforms.


styles/maxspeed.mapcss
```diff
@@ -64,9 +64,12 @@ way[!maxspeed].tracks
...
 
-way|z9-["construction:railway"="narrow_gauge"],
-way|z9-["construction:railway"="preserved"],
-way|z10-["construction:railway"="rail"]
+way|z9-["railway"="construction"]["construction:railway"="narrow_gauge"],
+way|z9-["railway"="construction"]["construction:railway"="preserved"],
+way|z10-["railway"="construction"]["construction:railway"="rail"],
+way|z9-["railway"="construction"]["construction"="narrow_gauge"],
+way|z9-["railway"="construction"]["construction"="preserved"],
+way|z10-["railway"="construction"]["construction"="rail"]
```
Nakaner: I wonder myself why we render narrow gauge railways earlier (zoom level 9) than other `construction=rail` (z10).

DerDakon:
I agree, but this is a different change. The decision when lines are shown is inconsistent at multiple places, see also #496.

PR #536, opened 2016-08-22
Improving Hp semaphore signal preset

Nakaner:
- add Zs 103
- add "no" as substitution signal because many semaphore main signals do
  not have a Zs 1


DerDakon:
Shouldn't the Zs103 preset link back to the main signal?

Nakaner:
Presets of Zs 2, Zs 2v, Zs 3 and Zs 3v don't link back to all main/combined signals. That's why I thought that it were not necessary.

PR #537, opened 2016-08-23
Fix wrong value of Ne 13

Nakaner:
So 18a/b has been forbidden a few years ago.


rurseekatze:
There were 4 signals using the deprecated tagging. I updated them in http://www.openstreetmap.org/changeset/42396771.

PR #545, opened 2016-09-14
validator: warn if ref contains both VzG and track number

DerDakon:


Nakaner:
I would prefer to use the warning instead of the error message type because JOSM's own validation rules rarely output errors – usually warnings only.

DerDakon:
We use errors for all other stuff that is clearly wrong, and AFAICT this is not only against the proposed tagging (although I don't find the place where I have read this anymore), but there is also no legal way to get this information.

PR #546, opened 2016-09-17
Add Finnish block, shunting, crossing and bridge protection signals

Nakaner:


styles/signals.mapcss
```diff
@@ -617,6 +630,44 @@ node|z17-[railway=signal]["railway:signal:direction"]["railway:signal:minor"="DE
...
+	text-halo-color: white;
+	font-weight: bold;
+	icon-image: "icons/fi/ro0-new-24.png";
+	icon-width: 11;
+	icon-height: 12;
+	text-allow-overlap: true;
+	allow-overlap: true;
+}
+
+/*****************************************************/
```
DerDakon: One * is missing here

styles/signals.mapcss
```diff
@@ -617,6 +630,44 @@ node|z17-[railway=signal]["railway:signal:direction"]["railway:signal:minor"="DE
...
+/* FI minor light signals type Lo at moveable bridges */
+/*****************************************************/
```
DerDakon: Another * missing.

Nakaner:
@DerDakon I added the missing *s.

styles/signals.mapcss
```diff
@@ -617,9 +630,47 @@ node|z17-[railway=signal]["railway:signal:direction"]["railway:signal:minor"="DE
 	allow-overlap: true;
 }
 
-/********************************************/
+/*******************************************/
+/* FI shunting light signals type Ro (new) */
+/*******************************************/
+node|z17-[railway=signal]["railway:signal:direction"]["railway:signal:shunting"="FI:Ro"]["railway:signal:shunting:states"="FI:Ro0;FI:Ro1;FI:Ro2"]["railway:signal:shunting:form"=light]
```
DerDakon: This enforces a specific ordering of the signal states, which we do not otherwise enforce because basically any order can happen. If exactly these 2 states must be present and nothing else then you should probably add another rule that checks for the opposite order, if additional states are permitted then you should use ~= and should check for both of them separately.

styles/signals.mapcss
```diff
@@ -617,9 +630,47 @@ node|z17-[railway=signal]["railway:signal:direction"]["railway:signal:minor"="DE
...
+	icon-width: 11;
+	icon-height: 12;
+	text-allow-overlap: true;
+	allow-overlap: true;
+}
+
+/******************************************************/
+/* FI minor light signals type Lo at moveable bridges */
+/******************************************************/
+node|z17-[railway=signal]["railway:signal:direction"]["railway:signal:minor"="FI:Lo"]["railway:signal:minor:states"="FI:Lo0;FI:Lo1"]["railway:signal:minor:form"=light]
```
DerDakon: The same applies here.

styles/signals.mapcss
```diff
@@ -1627,11 +1678,13 @@ node|z14-[railway=signal]["railway:signal:direction"]["railway:signal:main"="AT-
...
-/* FI main light signals */
-/* general rules         */
-/*************************/
-node|z14-[railway=signal]["railway:signal:direction"]["railway:signal:main"=~/^FI:Po(-v)?$/]["railway:signal:main:form"=light]
+/*****************************/
+/* FI main light signals     */
+/* FI combined block signals */
+/* general rules             */
+/*****************************/
+node|z14-[railway=signal]["railway:signal:direction"]["railway:signal:main"=~/^FI:Po(-v)?$/]["railway:signal:main:form"=light],
```
DerDakon: I think this section should be "inverted": for all 3 signals use only the matching rules below that match exactly one signal. End every of them using "set .fi-main-combined;" (or however you want to call that tag). Then move this section below the 3 others and change it to match "node.fi-main-combined", with the style rules as above. This will avoid doing the complete matches twice, and it will avoid the need for a regular expression match.

Nakaner:
@DerDakon I tried to adress all your complains.

DerDakon:
What about adding them to the legend?

DerDakon:
You reference eo2 old style in the legend, but it is not in the stylesheet.

Nakaner:
@DerDakon wrote:
> You reference eo2 old style in the legend, but it is not in the stylesheet.

I have removed Eo2 (old) from the legend.

Are there any other issues with this pull request?

rurseekatze:
Currently you are using retina icons in the legend, which means that the icon size of map and legend is not equal. You should also export the icons in their real size and correct the filenames.

styles/signals.json
```diff
@@ -260,6 +260,81 @@
...
+		},
+		{
+			"minzoom": 14,
+			"icon": "icons/fi/po0-new-30.png",
+			"caption": "Finnish main signal new type (aspects not given)"
+		},
+		{
+			"minzoom": 14,
+			"icon": "icons/fi/po1-new-30.png",
+			"caption": "Finnish main signal new type with Po&nbps;1"
```
DerDakon: This should be &nbsp;!

Nakaner:
I have added low-res icons and fixed the HTML entities.

rurseekatze:
Some test cases (not all added signals):

* **To** crossing signal (http://www.openrailwaymap.org/canvas.php?lat=60.291565876000476&lon=25.061209201812744&zoom=15&style=signals)
* **Ro** light shunting signal (http://www.openrailwaymap.org/canvas.php?lat=60.21044662563641&lon=24.92569863796234&zoom=17&style=signals)
* **Lo** minor light signal at bridges (http://www.openrailwaymap.org/canvas.php?&lat=59.98239198903981&lon=23.42737376689911&zoom=17&style=signals)

PR #561, opened 2016-11-19
fix errors in German tram signals preset

sb12:
I found two minor errors in the German tram signals preset:

- it should be `railway:signal:combined:states` for combined H-signals
- it should be `railway:signal:distant=DE-BOStrab:v` for distant H-signals


rurseekatze:
I also fixed wrong tagged signals in http://www.openstreetmap.org/changeset/44021545 and http://www.openstreetmap.org/changeset/44021612.

PR #580, opened 2017-02-27
add "electrified" style to available styles

mmoll:


Nakaner:
The `electrified` style is currently only provided as vector tiles but your pull request tries to add them to the web site where only raster tiles are used. Because our vector tiles have several bottlenecks (unsimplified geometries, no compression etc.), they are not advertised and the [web site where they are in use](http://www.openrailwaymap.org/canvas.php) is not linked from a well-known place (they are only used for style development).

Sorry but the chances that this pull request will be merged, are low.

mmoll:
@Nakaner Ah, I already thought there must be some reason (besides being overlooked historically). Thanks for the hint, I'll leave this just open for now as a reference.

DerDakon:
The other reason is that the prerendered bitmap tiles use a lot of disk space…

rurseekatze:
We can keep this pull request open, but it will take some time until we are ready to merge it. Before we can add more layers, there has to be more technical development for more efficient ressource usage.

Currently the vector layers scale very well to server many different styles, but their performance is too bad for productive usage. That's why the vector map is just unofficial and why we use bitmap tiles for our main website. But the disadvantage of bitmap tiles is that with every more layer you need more disk space, more rendering power, etc.

jonsger:
I have some questions here:

* Where are the bottle necks of the vector map and how could we tackle them?
* About which disk size do we talk here for an additional "electrified" layer? What do the existing layers need?
* Would it be an option to only offer some zoom levels? For me z14 or z15 would be enough as highest zoom level. 
* Is/are the server(s) to slow for rendering? Do we need to upgrade them hardware wise?

I would be very interested in an electrified layer. To see the state of electrification data in OSM and the general state of electrification :)

According to the [OSM wiki](https://wiki.openstreetmap.org/wiki/Tile_disk_usage) the disk size should be way below a TB, as most tiles on the planet doesn't have a railway on it :( If the money for hardware is the problem, let me know (now a days a 1TB M2 SSD costs 200 €)...

Nakaner:
The pull request will likely be merged in the next weeks. As the new owner of the project, I ordered a new server. It will have sufficient disk space to cache the tiles of four map styles. However, the new server uses a Mapnik based rendering stack, i.e. all styles need to be ported. This has been done for infrastructure, speed limits and signalling. However, electrification has a lower priority because my main priority is to set up the system and get it run more stable than now.

I started to port the electrification style (the current state of work so incomplete that I have not called `git push` yet). It seems to be less work than the signalling style. The repository of the Mapnik styles can be found at https://github.com/OpenRailwayMap/OpenRailwayMap-CartoCSS

@jonsger A tile cache for a transparent map style for zoom levels 0 to 12 (they are pre-rendered with the new technology) needs 10 GB for 256x256 pixel tiles. Retina tiles (512x512 tiles) need 2.5 times as much space, i.e. 25 GB. Because we offer more than zoom 0–12 and the number of tiles increases quadratically, it is safe to assume a cache size of 625 GB. The new server has two hard disks of 4 TB in a RAID 1 for tile cache and operation system.

Nakaner:
See https://github.com/OpenRailwayMap/OpenRailwayMap-CartoCSS/pull/24 for the progress of the work on the style.

DerDakon:
Since this seems to become a never ending story and since this is sort-of public anyway: how about just enabling this and labeling the style as "beta" for the time being?

mmoll:
> labeling the style as "beta" for the time being

I added it and hope this is how you had it in mind. :)

DerDakon (approved):

jonsger:
Thanks to all involved. It was mentioned on the homepage of the German railway NGO [ProBahn](https://www.pro-bahn.de/aktuell/index.php).
Screenshot_2021-02-14 Fahrgastverband PRO BAHN e V

PR #582, opened 2017-03-22
add a note regarding the osm2pgsql version

mmoll:


mmoll:
The packages provided by Debian 8 (without backports) and Ubuntu 16.04 (the latest LTS release) are not new enough for `nocolumn` in the import style.

PR #590, opened 2017-09-09
Add rendering GSM-R masts

Nakaner:
The icon lacks some creativity but it was fast to draw.

@bigbug21 might be interested in rendering of these masts.


bigbug21:
Thanks. It would be somewhat nice to see GSM-R masts on the map. In addition, it would be good if GSM-R antennas (such as on tunnel portals) would be shown as well. They are usually mapped with "man_made=antenna" and the usual GSM-R tags. (Perhaps this could be added to the templates as well)

Nakaner:
Dakon commented on the IRC that the icon looks more like a Lf 7 signal seen from its back side than a radio mast with antennas. That's why I drew a new icon.

Radio mast:

radio-mast-30

[Radio antenna](https://commons.wikimedia.org/wiki/File:GSMRRichtantennenFreudensteintunnelOst.jpg) as whished by @bigbug21:

radio-antenna-20

rurseekatze:
The new icons look much better. The old one really looks like the back side of a sign. Can you add the new icons to the style?

Nakaner:
I applied the changes, squashed the commits and rebased onto master.

Nakaner:
I rebased on current master and would appreciate a merge because I have been asked for progress on this issue.

rurseekatze:
You could also include the new icons in the JOSM preset.

rurseekatze:
The icon is not very readable because the underlying background map also renders communication masts with a similar icon:
screenshot at 2018-08-27 20 21 50

@Nakaner might it be better to replace the icon by [the one used in openstreetmap-carto](https://github.com/gravitystorm/openstreetmap-carto/blob/master/symbols/towers/communications_tower.svg)?

Nakaner:
@rurseekatze Or adding white rectangular casing similar to the `railway=phone` icon?

rurseekatze:
Yes, that is also a good solution. I think this would also be better readable than a thin black icon on a gray background. Please propose a version with some casing (round or rectangular?).

PR #592, opened 2017-10-31
Presets: add VBK Fv-Signals and improve VBK switch signals

sb12:
This assumes that the signal description at http://wiki.openstreetmap.org/wiki/DE:OpenRailwayMap/Tagging_Trams_in_Germany#Karlsruhe is correct.


sb12:
There's a Fv2 and Fv3 signal in the Oststadt: https://www.mapillary.com/map/im/8sKAGUk0AVqsRF6YzMYyBQ
The picture is from 2014, but it is still there.

PR #602, opened 2018-03-03
Web app manifest

rurseekatze:
Add web app manifest and mobile web app capability.


DerDakon:
I would suggest a more "speaking" name than manifest.json, e.g. webapp-manifest.json or webapp.json.

rurseekatze:
In https://www.w3.org/TR/appmanifest/ they use `manifest.webmanifest`.

DerDakon:
Also fine, although it will probably break automatic highligting in most editors.

rurseekatze:
I renamed it to `webapp-manifest.json`.

rurseekatze:
Yes, you can squash the commits together. Do you want to merge it manually instead of the Merge button in the UI?

The web app manifest makes the website feel a bit more like a real app. It allows to view the page without the browser frames and "install" it on the homescreen: https://developers.google.com/web/fundamentals/web-app-manifest/. But AFAIK it only works on Chrome for Android and Firefox Mobile.

rurseekatze:
Merged.

PR #603, opened 2018-03-03
Update locales

rurseekatze:
* Updates translations from Transifex, update Authors list
* Remove translation strings no longer needed
* Sync translation strings between languages
* Add new translation strings
* Add new languages:
  * Catalan `ca_ES`
  * Chinese Taiwan `zh_TW`
  * Finnish `fi_FI`
  * Japanese `ja_JP`
  * Lithuanian `lt_LT`

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,7 +1,7 @@
 Contributors
 ------------
 
-Alexander Matheisen (rurseekatze) <AlexanderMatheisen at ish.de>
+Alexander Matheisen (rurseekatze) <alex@matheisen.org>
  - Project founder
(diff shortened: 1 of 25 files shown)


DerDakon:
The German localization looks sane. I wonder about the "Last-Translator" that always points to you, even if you likely did not any real translations. I guess that is a result of your tooling?

rurseekatze:
I also noticed that Last-Translator often mentions me. This is how Transifex exports the files. But in the past there were more files with me as Last-Translator, in some files I disappeared in the latest export.

PR #604, opened 2018-03-04
Added client side i18n, remove timestamp API and generate last update date on client side

rurseekatze:
A summary of the changes:
* Convert gettext .po files to JSON with the tool pojson, so that they can be used in clientside code, remove jstranslations.php
* Remove timestamp.php, fetch directly the timestamp file and generate the human-readble string on client side

This branch is checked out on https://dev.openrailwaymap.org/, you can test the code.


DerDakon:
The timestamp age is not properly shown for Danish and Spanish.

rurseekatze:
First part: https://github.com/OpenRailwayMap/OpenRailwayMap/pull/605

rurseekatze:
Next: https://github.com/OpenRailwayMap/OpenRailwayMap/pull/606

rurseekatze:
Next: https://github.com/OpenRailwayMap/OpenRailwayMap/pull/607

rurseekatze:
A solution to fix the CI run would be upgrading to Python 3.3. https://docs.python.org/3/library/stdtypes.html#text-sequence-type-str says:

> Changed in version 3.3: For backwards compatibility with the Python 2 series, the u prefix is once again permitted on string literals. It has no effect on the meaning of string literals and cannot be combined with the r prefix.

DerDakon:
You need to squash https://github.com/DerDakon/OpenRailwayMap/commit/ad33c11a4e44edd5885084343f3849619c4e92fc into the commit that adds pojson to make the CI work again.

rurseekatze:
I increased the Python version in https://github.com/OpenRailwayMap/OpenRailwayMap/pull/604/commits/ae6cfc2ab39718efb114937acf8248bfd9612001, now the CI run is successful.

js/functions.js
```diff
@@ -6,6 +6,31 @@ See http://wiki.openstreetmap.org/wiki/OpenRailwayMap for details.
...
+	"pt": "pt_PT",
+	"ru": "ru_RU",
+	"sl": "sl_SI",
+	"sv": "sv_SE",
+	"tr": "tr_TR",
+	"uk": "uk_UA",
+	"vi": "vi_VN",
+	"zh": "zh_TW"
+};
+
```
DerDakon: inconsistent newline count: either 1 as here or 2 as above the function.

js/functions.js
```diff
@@ -223,21 +248,34 @@ function setupControls()
...
+function _(sourcemsg, n)
+{
+	if (typeof translations[sourcemsg] === 'undefined')
+		return sourcemsg;
+
+	if (n === undefined)
+		return translations[sourcemsg][1];
+
+	var pluralForms = translations[""]["Plural-Forms"].match(/plural=(.*);/)[1];
```
DerDakon: This is a bit too simplified to properly handle even western European languages, but it is good enough for German and English. Can always be improved later (preferably by reusing other peoples code).

js/bitmap-map.js
```diff
@@ -21,6 +21,7 @@ function createMap(embed)
 		"signals": "Signalling"
 	};
 
+	translations = null;
```
DerDakon: Should be squashed into the commit that introduced the bug.

js/bitmap-map.js
```diff
@@ -110,7 +108,7 @@ function setStyle(style)
 	// helper variable for saving current map style
 	railmap.selectedStyle = style;
 	// change tileserver url to load different style
-	railmap._url = tiledir+style+'/{z}/{x}/{y}.png';
+	railmap._url = window.openrailwaymap.tiledir+style+'/{z}/{x}/{y}.png';
```
DerDakon: when you touch the line anyway, how about some spaces around the operators?

js/bitmap-map.js
```diff
@@ -14,6 +14,30 @@ window.openrailwaymap = {
 		"standard": "Infrastructure",
 		"maxspeed": "Maxspeeds",
 		"signals": "Signalling",
+	},
```
DerDakon: Please squash this into the previous commit (or even earlier, depends on which introduced the actual bug).

js/functions.js
```diff
@@ -89,7 +89,7 @@ function getRequest(url, handler)
 
 	request.onerror = function()
 	{
-		console.log('Error: ' + error);
+		console.log('Error: ' + request.status);
```
DerDakon: squash

js/timestamp.js
```diff
@@ -45,10 +45,10 @@ function Timestamp(box)
 			{
 				var timestamp = parseInt(response);
 				var lastUpdate = new Date(0);
-				lastUpdate.setUTCSeconds(timestamp * 1000);
+				lastUpdate.setUTCSeconds(timestamp);
```
DerDakon: squash

api/functions.php
```diff
@@ -250,7 +250,7 @@ function urlArgsToParam($checkMobile, $urlbase)
 				else
 					echo "null,\n";
 			echo "zoom : " . (isValidZoom('zoom') ? ($_GET['zoom']) : ("null")) . ",\n";
-			echo "lang : " . (isset($_GET['lang']) ? ($_GET['lang']) : ("null")) . ",\n";
+			echo "lang : " . (isset($_GET['lang']) ? ("'".$_GET['lang']."'") : ("null")) . ",\n";
```
DerDakon: squash into previous

js/timestamp.js
rurseekatze: Squashed.

api/functions.php
rurseekatze: Squashed.

js/functions.js
rurseekatze: What is the problem with this code? In my opinion it works for all languages we currently have in the repository.

js/bitmap-map.js
rurseekatze: Instead of fixing coding style in this pull request, I would propose to do this later by integrating a tool like ESLint in out workflow.

js/functions.js
rurseekatze: Squash with which commit?

js/bitmap-map.js
rurseekatze: What is wrong here?

js/bitmap-map.js
rurseekatze: Squashed.

js/bitmap-map.js
DerDakon: Possible. I tend to fix such things in lines I touch anyway. YMMV.

js/functions.js
DerDakon: Seems like an older bug, so keep it as is.

PR #608, opened 2018-03-04
RSS feed link

rurseekatze:
Fix URL to feed and offer both languages to the user.

### DIFF
--- a/canvas.php
+++ b/canvas.php
@@ -31,8 +31,8 @@
 		<meta name="page-topic" content="OpenRailwayMap" />
 		<meta name="robots" content="index,follow" />
 
-		<link rel="alternate" type="application/rss+xml" title="OpenRailwayMap Deutsch" href="http://blog.openrailwaymap.org/de.rss" />
-		<link rel="alternate" type="application/rss+xml" title="OpenRailwayMap English" href="http://blog.openrailwaymap.org/en.rss" />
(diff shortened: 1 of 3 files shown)


canvas.php
```diff
@@ -30,7 +30,10 @@
 		<meta name="date" content="2010-01-01" />
 		<meta name="page-topic" content="OpenRailwayMap" />
 		<meta name="robots" content="index,follow" />
-		<link rel="alternate" type="application/rss+xml" title="OpenRailwayMap RSS Feed" href="http://www.matheisen.org/<?php echo ($lang == 'de') ? 'de' : 'en'; ?>/orm.rss" />
+
+		<link rel="alternate" type="application/rss+xml" title="OpenRailwayMap Deutsch" href="http://blog.openrailwaymap.org/de.rss" />
+		<link rel="alternate" type="application/rss+xml" title="OpenRailwayMap English" href="http://blog.openrailwaymap.org/en.rss" />
```
Nakaner: Could you please add `hreflang=de` and `hreflang=en`?

Could you either use HTTPS or let the client decide which protocol to use? `href=//blog.openrailwaymap.org/de.rss`

https://support.google.com/webmasters/answer/189077?hl=de

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#attr-hreflang

PR #610, opened 2018-03-05
Add webfont for icons

rurseekatze:
Add custom webfont generated with http://fontello.com/. It will be used for icons on a future version of the website. The included `font/config.json` can be used to re-generate the custom font. Some CSS files mentioned in `font/README.txt` were removed because they are not necessary here.

### DIFF
--- a/font/LICENSE.txt
+++ b/font/LICENSE.txt
@@ -6,7 +6,7 @@ Font license info
    Copyright (C) 2016 by Dave Gandy
 
    Author:    Dave Gandy
-   License:   SIL ()
+   License:   SIL (http://scripts.sil.org/OFL)
    Homepage:  http://fortawesome.github.com/Font-Awesome/
 
 


Nakaner:
If the code is needed for the new version of the web frontend, why don't you create a separate branch for the new frontend and file the pull request *against* that branch?

font/LICENSE.txt
```diff
@@ -0,0 +1,12 @@
+Font license info
+
+
+## Font Awesome
+
+   Copyright (C) 2016 by Dave Gandy
+
+   Author:    Dave Gandy
+   License:   SIL ()
```
Nakaner: Please add either the link to the license document or the text of the license itself.

rurseekatze:
> If the code is needed for the new version of the web frontend, why don't you create a separate branch for the new frontend and file the pull request against that branch?

I do not understand the problem. I could have created one big pull request for the website redesign. But I wanted to split up the work, so I decided to make a separate pull request for the font. Because the website depends on the font, it does make sense to add the font to the repository before I will create pull requests for the website redesign.

font/LICENSE.txt
rurseekatze: OK, I added a link to the license.

Fontello says on https://github.com/fontello/fontello/wiki/What-about-license%3F:

> If you place content of fontello's archive somewhere with your source code, that's enough. See, example for fontello itself.

DerDakon:
I think this boils down to: this is not needed as it is not used. You could just leave this open until you have the first usage of it ready, this should not cause any collisions so merging it later will likely not cause any problems.

PR #612, opened 2018-03-05
Add facebook opengraph tags

rurseekatze:


DerDakon:
Please provide a preview of what this causes.

rurseekatze:
A preview is not possible as Facebook requires to log in to use https://developers.facebook.com/tools/debug/. Bye bye, Facebook!

PR #616, opened 2018-04-06
Fix timestamp: use seconds instead of milliseconds

rurseekatze:


DerDakon:
do the substraction first, divide only once

DerDakon:
See [my branch](https://github.com/DerDakon/OpenRailwayMap/tree/timestamp-fixes) for my fixes.

rurseekatze:
I replaced my changes by your branch.

Issue #627, opened 2018-08-05
add support for station=subway


DerDakon:
Subway stations are also tagged as railway=station, but have an additional [station=subway](https://wiki.openstreetmap.org/wiki/Tag:station%3Dsubway) tag. Supporting this could make things more clear.

Things that need to be changed:

- [ ] rendering
- [ ] search API (server side)
- [ ] search API (client side) js/search.js and translation
- [ ] validator?
- [ ] presets?

PR #629, opened 2018-08-06
standard style: handle subway and light_rail stations separately

DerDakon:
Handles the first part of #627.

### DIFF
--- a/locales/ca_ES/LC_MESSAGES/messages.po
+++ b/locales/ca_ES/LC_MESSAGES/messages.po
@@ -176,6 +176,9 @@ msgstr "Lloc"
 msgid "Service station"
 msgstr "Estació de servei"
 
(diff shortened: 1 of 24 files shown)


rurseekatze:
The code is ok, but I wonder if it is a good idea to use cursive font to distinguish between train stations and subway/light rail stations. Cursive font 

Maybe it is better to use non-cursive and non-bold font and use the font color to distinguish the type of stations. Light rail stations could be rendered with green font and subway stations with blue font. In that case it would be necessary to change the font color of train stations to black.

DerDakon:
I decided against that because green text has rather low contrast, and is probably unreadable for some people.

DerDakon:
ping

DerDakon:
I messed it up…

PR #632, opened 2018-08-23
Polish JOSM presets

gscscnd:
This is an initial (yet mostly complete) version of two JOSM presets (_pl_ and _pl-signals_).


josm-presets/pl-signals.xml
```diff
@@ -0,0 +1,717 @@
...
+			<item name="Semafor kształtowy jednoramienny" type="node">
+				<key key="railway:signal:main" value="PL-PKP:sr"/>
+				<key key="railway:signal:main:form" value="semaphore"/>
+				<key key="railway:signal:main:states" value="PL-PKP:sr1;PL-PKP:sr2"/>
+				<reference ref="signal_ref"/>
+				<check key="railway:signal:main:deactivated" text="Wyłączony z użytku"/>
+				<reference ref="signal_position"/>
+				<reference ref="signal_directions"/>
+				<space/>
+				<combo display_values="normalny" key="railway:signal:main:height" text="Wysokość [m]" values="normal"/>
```
DerDakon: Do these signals only exist in one height? If yes, is a combobox really the right input for that? Other options would be e.g. key (i.e. always write that when the preset is used) or maybe a checkbox. Maybe also move this into a chunkß

gscscnd: AFAIK, Polish mechanical semaphores either have “normal” height or are mounted above tracks. I don’t know if they’re always of the same exact height, so I’ve used a combobox to allow specifying exact height if it’s known. What do you think about replacing it with
```xml
<key key="railway:signal:main:height" match="key" value="normal"/>
```
?

gscscnd:
How should XML attributes be ordered in `de.xml`? In some places `de.something` is set before `something` because `'d' < 's'` and in other `de.something` goes after `something` because it’s `something`’s translation (not to mention that there’s no clear order at all in most of the code). Cleaning it up isn’t the main point of this PR yet I’d prefer to not make it worse.

DerDakon:
JOSM will sort the entries according to their translation if values_sort is not set, so if that is not given the order does not matter. The order of attributes to a given key doesn't matter either.

More details can be found in the [JOSM wiki](https://josm.openstreetmap.de/wiki/TaggingPresets).

DerDakon:
Does that tagging match what has been described in the [OSM wiki](https://wiki.openstreetmap.org/wiki/DE:OpenRailwayMap/Tagging_in_Poland)? Can you please translate that page to Polish, I guess there will be the need to read about that tagging once it comes into usage.

DerDakon:
@gszy ping

gszy:
I have had translated most of the Wiki page, but still have to rethink some parts, like the light main and combined signals. This PR is far from ready to be merged and I am a bit short on time these days, so I would prefer to add the presets in smaller steps, starting from signs and semaphores, since I think their tagging schemes are in quite good shape. I will rebase this PR on my current files once I clean them up a bit, hopefully in September, and send other contributions in future PRs.

DerDakon:
I'll gladly accept a MR that adds Polish translations to the elements already in de.xml, i.e. no new features but only strings.

PR #643, opened 2018-10-28
Czech translations of maxspeeds, electrification and general

dbrvnistan:
Skipped translations of signalling as source is mainly in German and I am expert at neither.

Not sure about context / meaning of this one:
```
msgid "contact line"
msgstr "Trolejové vedení"
```
If catenary wire is meant, than this is correct, else I would need context.


DerDakon:
Yes, that means that this track is electrified with an overhead wire.

dbrvnistan:
> Yes, that means that this track is electrified with an overhead wire.

Thanks for confirmation!

PR #644, opened 2018-10-28
English corrections and additions

dbrvnistan:
Added missing "translations" to English.
I could not do de>en translations.
Is the msgid (German in some cases) value used in case msgstr is empty?

Corrected some terms to match the correct spelling or most widely used term from several choices e g. these two that you may disagree with:
highspeed > high-speed (https://en.wikipedia.org/wiki/High-speed_rail)
subway > metro (https://en.wikipedia.org/wiki/Rapid_transit#Terminology)

### DIFF
--- a/locales/en_GB/LC_MESSAGES/messages.po
+++ b/locales/en_GB/LC_MESSAGES/messages.po
@@ -22,80 +22,80 @@ msgstr ""
 "X-Poedit-SourceCharset: utf-8\n"
 
 msgid "More Info"
-msgstr "More Info"
+msgstr ""
 
 msgid "Contact"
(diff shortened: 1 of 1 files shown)


DerDakon:
If there is no translation then the original string is used.

Nakaner:
Thank you @dbrvnistan for your work. I discussed this with @DerDakon on IRC today. Your improvements to the translations are fine but we think that it duplicating the source strings is not bad. It prevents others from trying to "translate" them in future and saves their time.

I don't agree with your change of the translation of "imprint". It should still be called "imprint" for legal reasons (German law). Therefore, I am closing this pull request now. It is #666.

PR #654, opened 2019-04-20
Use Python 3.3 compatible packages to fix travis run

mmoll:


mmoll:
The alternative would be to update Python to at least 3.5.

DerDakon:
Given that python <3.5 is not supported by upstream anymore I would have said kill it, but the CentOS on our server is still at 3.4, so we better keep that working.

PR #669, opened 2019-11-23
Leaflet: Use released archives from GitHub and extract required files

Nakaner:
cdn.leafletjs.com is hosted by Cloudfront but has certificate errors (its certificate is valid for cloudfront.net and *.cloudfront.net only). The Leaflet website links to these archives without SSL only but I think that instead of giving up SSL, we should work with the released ZIP archive from GitHub.

The file from GitHub contains the whole repository, not the release only. Therefore, the extraction of the archive is a bit more difficult.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,28 +28,22 @@ check:
 distclean: clean
 	rm -rf download css/leaflet.css
 
-download/leaflet-$(DEP_LEAFLET_VERSION).zip:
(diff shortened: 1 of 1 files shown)


DerDakon:
First thing is: when you pick it from GitHub, then just take the .tar.gz, which can even be filtered on extraction. And then we will look on what problems remain ;)

PR #672, opened 2019-11-26
Set width explictly to get lines coloured in JOSM

Nakaner:
The default styling for tracks without train protection information was removed in #550 because JOSM users usually load multiple styles.

JOSM 15492 does not render lines without width set.

Changes for the other styles will be added later, that's why it is a draft pull request.

screenshot1
screenshot2

top: before change, bottom: after change


Nakaner:
The error does not occur if another style is loaded which sets the width for the railway tracks.

PR #681, opened 2020-02-10
dismanteld = razed

ferdinand0101:
according to the wiki dismanteld  is equal to razed for railways this pull request changes the validator to support dismanteld:railway
and changes the style to show dismanteld railways the same as razed railways

### DIFF
--- a/styles/common.mapcss
+++ b/styles/common.mapcss
@@ -47,6 +47,12 @@ way|z9-[railway=razed]["razed:usage"="main"][!"razed:service"]["razed:railway"!=
 way|z9-[railway=razed]["usage"="main"][!"service"]["razed:railway"!="light_rail"]["razed:railway"!="subway"]["razed:railway"!="tram"],
 way|z10-[railway=razed]["razed:railway"!="tram"],
 way|z11-[railway=razed]["razed:railway"="tram"]
+way|z9-[railway=dismantled]["dismantled:usage"="branch"][!"dismantled:service"]["dismantled:railway"!="light_rail"]["dismantled:railway"!="subway"]["dismantled:railway"!="tram"],
+way|z9-[railway=dismantled]["usage"="branch"][!"service"]["dismantled:railway"!="light_rail"]["dismantled:railway"!="subway"]["dismantled:railway"!="tram"],
+way|z9-[railway=dismantled]["dismantled:usage"="main"][!"dismantled:service"]["dismantled:railway"!="light_rail"]["dismantled:railway"!="subway"]["dismantled:railway"!="tram"],
+way|z9-[railway=dismantled]["usage"="main"][!"service"]["dismantled:railway"!="light_rail"]["dismantled:railway"!="subway"]["dismantled:railway"!="tram"],
+way|z10-[railway=dismantled]["dismantled:railway"!="tram"],
+way|z11-[railway=dismantled]["dismantled:railway"="tram"]
 {
 	set .tracks_razed;
 }
--- a/styles/standard.mapcss
+++ b/styles/standard.mapcss
@@ -80,6 +80,7 @@ way|z12-[!"railway:track_ref"].bridge::bridges
 way|z9-[railway=disused].tracks,
 way|z9-[railway=abandoned].tracks,
 way|z9-[railway=razed].tracks
+way|z9-[railway=dismantled].tracks
 {
 	z-index: 10;
 	casing-color: black;
@@ -154,6 +155,11 @@ way|z11-[railway=razed][!bridge][!tunnel]["razed:name"],
 way|z11-[railway=razed][!bridge][!tunnel]["razed:ref"]
 {
 	text: eval(join(" ", tag("razed:ref"), tag("razed:name")));
+}
+way|z11-[railway=dismantled][!bridge][!tunnel]["dismantled:name"],
+way|z11-[railway=dismantled][!bridge][!tunnel]["dismantled:ref"]
+{
+	text: eval(join(" ", tag("dismantled:ref"), tag("dismantled:name")));
 }
 
 /*****************************************/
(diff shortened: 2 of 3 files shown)


styles/standard.json
```diff
@@ -269,6 +269,14 @@
 					"razed:railway":"rail",
 					"railway":"razed"
 				}
+			}
+		{
```
DerDakon: indentation error

styles/standard.mapcss
```diff
@@ -80,6 +80,7 @@ way|z12-[!"railway:track_ref"].bridge::bridges
 way|z9-[railway=disused].tracks,
 way|z9-[railway=abandoned].tracks,
 way|z9-[railway=razed].tracks
```
DerDakon: comma missing, this will make everything not work anymore

styles/standard.mapcss
```diff
@@ -464,6 +470,7 @@ way|z9-[railway=abandoned]
 /* razed railways */
 /******************/
 way|z9-[railway=razed]
```
DerDakon: another comma missing

ferdinand0101:
https://wiki.openstreetmap.org/wiki/Railways#Life-cycle

> Obliterated (or dismantled/razed) - Some people use railway=dismantled or railway=razed where all evidence of the line has been removed, others prefer such lines not to be included at all. This is used where the alignment has been replaced by new buildings or by roads which don't refer to the old alignment. Alternatively where the alignment now crosses farmland.

styles/standard.json
ferdinand0101: I still can't figure that part out it has me confused

ferdinand0101: Okay I managed to fix it

styles/standard.json
```diff
@@ -268,8 +268,15 @@
...
-			}],
+					
+				}},					
+				{"type": "LineString",
+				"coordinates": [[10,50],[60,50]],
+				"properties": {
+					"dismantled:railway":"rail",
+					"railway":"dismantled"
+				}}			
+			],
```
Nakaner: You don't need to duplicate this legend entry. Both variations of tagging (razed and dismantled) are rendered the same. The way you currently did it, leads to to lines at the same location above each other. It has no negative effect on rendering the legend but it is unnecessary.

ferdinand0101: > razed and dismantled are rendered the same.
Really I thought before this change dismanteld would not be renderd

Nakaner:
I dug a bit through the wiki and mailing list archives to answer myself the question whether `railway=razed` or `railway=dismantled` is better.

From 2011 to end of 2014, `dismantled` [dominated](https://taghistory.raifer.tech/#***/railway/dismantled&***/railway/razed). Since then, `razed` has dominated. I presume that that is the fault of OpenRailwayMap.

There was a discussion about "strange tags for traffic routes" in June 2013 which was about proposed roads and abandoned/razed/… railways. In one posting, a user [mentions](https://forum.openstreetmap.org/viewtopic.php?pid=338827#p338827)  that the wiki was confusing back then. That was in June 2013. At end of June  2013, someone created [Demolished_Railway](https://wiki.openstreetmap.org/wiki/Demolished_Railway) which offers the different ways to tag it.

The wiki page [DE:Railways](https://wiki.openstreetmap.org/wiki/DE:Railways) has mentioned dismantled only in the beginning (created in April 2013 by translation), razed was [added](https://wiki.openstreetmap.org/w/index.php?title=DE%3ARailways&type=revision&diff=1430389&oldid=1401426) by myself two years ago.

`razed` was https://wiki.openstreetmap.org/w/index.php?title=Railways&type=revision&diff=1066574&oldid=1066571 to [Railways]() in July 2014. Before that, it mentioned `dismantled` only.

The talk page Talk:Railways does not contain any helpful information.

The page Tage:railway=dismantled was created in June 2013 as the predecessor of Demolished_Railway (later renamed to Demolished_Railway).

Mailing lists do not provide any help. They only discussed whether it should be mapped at all, not which tag to use.

ferdinand0101:
so what should we do just close it ?

Nakaner:
> so what should we do just close it ?

No, not now. I wait for a comment by the founder of the project what he had in mind regarding razed vs. dismantled.

We do not have to fully close and reject your pull request. I am in favour of a basic support of `railway=dismantled` (i.e. no support of `dismantled:railway=*` which is used 8 times only).

styles/standard.json
ferdinand0101: okay correction: so I copied this scheme from the main rail lines which alsohave multiple ways of being tagged.

Isn't the "duplicate" needed for if there is no razed but only dismanteled in an area so it still shows up in the legend

styles/standard.mapcss
```diff
@@ -155,6 +155,11 @@ way|z11-[railway=razed][!bridge][!tunnel]["razed:ref"]
 {
 	text: eval(join(" ", tag("razed:ref"), tag("razed:name")));
 }
+way|z11-[railway=dismantled][!bridge][!tunnel]["dismantled:name"],
```
DerDakon, line 158: I don't think this is worth it. There are as of today 6 ways with dismantled:name and none with dismantled:ref in taginfo.

styles/common.mapcss
```diff
@@ -47,6 +47,12 @@ way|z9-[railway=razed]["razed:usage"="main"][!"razed:service"]["razed:railway"!=
 way|z9-[railway=razed]["usage"="main"][!"service"]["razed:railway"!="light_rail"]["razed:railway"!="subway"]["razed:railway"!="tram"],
 way|z10-[railway=razed]["razed:railway"!="tram"],
 way|z11-[railway=razed]["razed:railway"="tram"]
+way|z9-[railway=dismantled]["dismantled:usage"="branch"][!"dismantled:service"]["dismantled:railway"!="light_rail"]["dismantled:railway"!="subway"]["dismantled:railway"!="tram"],
```
DerDakon, line 50: I don't think this is worth it, dismantled:railway is barely used and dismantled:usage is never used, so just scrap that.

DerDakon (changes requested):

DerDakon:
@Nakaner do we just want to start this as a new PR and just drop out the parts we don't need?

PR #683, opened 2020-03-09
Update RU locale

freeExec:
Translated the remaining lines


locales/ru_RU/LC_MESSAGES/messages.po
```diff
@@ -451,103 +452,103 @@ msgid "Blog"
...
-msgstr ""
+msgstr "Ответвляющийся узел"
 
 msgid "No background map"
-msgstr ""
+msgstr "Нет подложки карты"
 
 msgid "Deelectrified"
-msgstr ""
+msgstr "Деэлектрификация"
```
mmoll: I'm not a native speaker, but wouldn't Деэлектрифицированный be a more direct translation?

freeExec: You right, thanks.

DerDakon:
So it looks like @mmoll thinks it's somewhat sensible, and I can't judge, so I would just go ahead and merge this if noone objects.

PR #701, opened 2021-01-18
Added support for Dutch ETCS block markers

JJJWegdam:
- edited signals.mapcss
- added etcs-stop-marker-triangle-left.svg to icons
- added etcs-stop-marker-triangle-left-32.png to icons

etcs-stop-marker-triangle-left-32


DerDakon:
According to the German Wikipedia this is an older version of the signal which has been replaced in newer versions of the ETCS standard because it could be confused with a France signal of different meaning. Will the existing signals in NL be upgraded to the new drawing?

JJJWegdam:
The Netherlands has two ETCS level 2 trajectories (railway lines with ETCS block markers). The high speed line between Amsterdam and Antwerp has (both on Dutch and Belgian soil) the triangle-shaped signs. The cargo line from Rotterdam to the Ruhrgebiet area has arrow-shaped signs. **There are no plans to change the triangle-shaped block marker boards** on the high speed line. The current situation seems to make sense to me: the railway line to Germany uses the German standard and the railway line to Belgium/France uses the French standard.

The Dutch national law (https://wetten.overheid.nl/BWBR0017707/2020-04-01 appendix 2, chapter 13) literally allows both signs under the same signal reference number:
image

This makes tagging a bit hard, because `"railway:signal:train_protection"="NL:227b"` can mean both a triangle-shaped or an arrow-shaped sign (depending on the railway track you're visiting). My current solution is to use `"railway:signal:train_protection"="DE-ESO:ne14"`on the Germany-bound line and  `"railway:signal:train_protection"="NL:227b"` on the Belgium-bound line. This is not ideal, so if you know a better way of adressing this situation feel free. I'll happily change the tags and pull request. In the end the goal is to render everything, the way to get there is flexible for me.

DerDakon:
I guess this is similar to the at least 3 different versions of H/V light signals in Germany, that are all tagged the same. I think this is a tagging issue in the first place _if_ we want to address this. But I think this is the wrong place to discuss the general tagging. This should be brought up on the mailing list, and when a conclusion is reached then the tagging should be set here.

JJJWegdam:
I agree. We have the choice for two solutions:
1. proceed with this PR, tag the German-bound line with `"railway:signal:train_protection"="NL:227b"` and be done with it
2. close this PR, discuss on the mailing list, come up with a new tagging solution and reopening an improved PR

My preference is option 1. I hope you agree

DerDakon:
Any tagging of DE-ESO signals on a railway line in the Netherlands is plain wrong, this is just "tagging for the renderer". We can merge this as is, I have no strong opinion on this.

JJJWegdam:
Fair point. Tagging of DE-ESO signals in the Netherlands is indeed not a solution. I will:

- close this PR
- write a tagging proposal for the mailing list
- if this is approved, change the wiki and branch
- reopen the PR

PR #703, opened 2021-01-26
Add EBICAB captions to track signaling.

HedgiePT:


JJJWegdam:
_This is relevant for this PR so I added it here. This is a tagging discussion so I added a copy to the OpenRailwayMap mailing list._

Great to see an initiative to render an additional train protection system. Very sorry to delay it, but I think we first have to make a broader decision.

EBICAB is a trademark for on-board equipment, from a specific supplier (Bombardier). The entire train protection system contains some other things [1]. The entire system is called ATC in Norway and Sweden, while Portugal calls the exact same system CONVEL. To add more confusion: Denmark calls [its own system](https://de.wikipedia.org/w/index.php?title=ZUB_123) ATC [2], while it is incompatible with the Norse/Swedish/Portuguese system.

Also the current situation in OSM is different than you currently envision (with adding the `railway:ebicab=700` tag): in the past I already added `railway:atc=yes` tags to relevant tracks in both Portugal and Norway with the same purpose. You also envision the `railway:ebicab=900` tag (probably for Finland) while Finland uses the `railway:jkv=yes` tag with the same purpose. Denmark is a bit of a blank slate, because Denmark doesn't have train protection tags yet.

We have 2 options:

1. Render compatible systems
   consequences:
   a) we proceed with this PR as is
   b) we have to retag Portugal, Sweden, Norway and Finland (I'm willing to help)
   c) we should create an additional PR to also render `railway:zub=123`
   d) Denmark should be tagged with `railway:zub=123`
2. Render local names
   consequences
   a) this PR should change to `railway:convel=yes`
   b) we should retag portugal from `railway:ebicab=700`/`railway:atc=yes` to `railway:convel=yes` (I'm willing to help)
   c) we need an additional PR to render `railway:atc=yes`
   c) Denmark should be tagged with `railway:atc=yes`
   
I am in favor of option 1, because my opinion is that the goal of ORM signalling layer should be to show compatibility.

[1] Overview of train protection systems in PT, DK, NO, SE, FI as far as I currently understand them 
image

[2] Danish border, seen from Germany. Note the start-of-ATC signs.
image

RossGammon:
Hi,
I came across this PR after wondering why my tagging of one line section with railway:zub123 (as described in https://wiki.openstreetmap.org/wiki/OpenRailwayMap/Tagging#Train_protection_systems) was not rendering. It is clear now that a PR is probably required!
I think the previous comment (which is a probably a bit wider than the scope of this PR) brings up an interesting question about the tagging of train protection systems in general.
In the Nordic example, the designation Automatic Train Control (ATC) is widely used on signposts and in the operational rule books and instructions. But technically, these systems are really Automatic Train Protection (ATP) systems (although there is an element of control in that the protection limits are communicated to the driver and can be dynamically updated).
It is generally accepted that an ATC system is made up of three subsystems: Automatic Train Supervision (ATS), Automatic Train Operation (ATO), and ATP.
In the case of Denmark, there is HKT on the Copenhagen suburban lines, ATC (Siemens ZUB123) on the mainlines, and ATC (Hitachi full driverless ATC system with MicroLok ATP & ATO). There is no tag on the wiki for the last one.
Maybe we need additional tags for "local_name" and another for "system_type"?
But with respect to this PR I would vote to render the systems as proposed (Ebicab 700/900), as it is interesting that the same family of ATP system is used in multiple countries.

DerDakon:
Tagging should be discussed on the mailinglist, please use PRs only for discussions of technical aspects.

JJJWegdam:
A discussion on the mailing was already started last Saturday @DerDakon. @RossGammon feel free to re-post your message there

JJJWegdam:
The discussion on the mailing list has not been answered for one week now. I would estimate that people agree with the proposed 'option 1' and thus that this PR can be reviewed/merged. Would you agree @DerDakon @Nakaner ?

PR #705, opened 2021-02-03
Signals4denmark

RossGammon:
It would be nice to see the results of tagging of the ATC system in Denmark on Open Railway Map.

As there is a programme to completely replace ATC with ETCS over the coming ~10 years, it would be nice to show the progress over time.

### DIFF
--- a/styles/signals.json
+++ b/styles/signals.json
@@ -100,6 +100,58 @@
 			"caption": "ETCS under construction"
 		},
 		{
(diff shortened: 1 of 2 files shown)


DerDakon:
Please bear in mind that these style are only used for JOSM and the legend, what is actually shown in the map is in the repo OpenRailwayMap-CartoCSS.

PR #706, opened 2021-02-17
Move PHP files out of api directory

mmoll:


mmoll:
The legend generator is responding with HTML, not an API response, so IMHO this is reasonable.

This would open up the path to extract the current NodeJS API into its own repo and also to seperate out its deployment. I have the beginning of surface testing the API in my branch `api_tests` here: https://github.com/mmoll/OpenRailwayMap/actions/runs/572702320

All this would make the API deployable standalone or also containerized and also facilitate the future switch to i.e. the Python rewrite.

mmoll:
I just found a fix locally for the non-working locales also, but it would require all the PHP files to be in the same directory...

Nakaner (approved):

DerDakon:
So, does this still need any additional things to get the localization working? Or is everything ready?

mmoll:
Everything is ready. :)

PR #707, opened 2021-02-17
Add Leaflet.EditInOSM files to .gitignore

mmoll:


mmoll:
An alternative would be to have a gitignored `3rdparty` or `vendor_assets` (or similar) dir for files that are generated on deployment.

DerDakon (approved):

PR #708, opened 2021-02-18
Rename electrified -> electrification

mmoll:


mmoll:
At some point in time this started to diverge. In the new stack it's "electrification", so unify it here. This should fix the legend on the live instance.

I wonder how mod_tile and tirex are configured in order for that part to work currently, however, a ansible run with `--diff` would tell...

Nakaner:
I thought about renaming when I merged and deployed #580. I decided against it for the time being (and added a redirect in Apache for the map tiles) because it breaks for those how use the style in JOSM.

According to yesterday's (Friday) Apache logs, all the MapCSS map styles get queried for about 9 to 17 times per day by JOSM:

```sh
michael@buegelfalte:~$ sudo grep standard.zip /var/log/apache2/www.openrailwaymap.org.access.log.1 | grep JOSM | wc -l
15
michael@buegelfalte:~$ sudo grep signals.zip /var/log/apache2/www.openrailwaymap.org.access.log.1 | grep JOSM | wc -l
17
michael@buegelfalte:~$ sudo grep maxspeed.zip /var/log/apache2/www.openrailwaymap.org.access.log.1 | grep JOSM | wc -l
9
michael@buegelfalte:~$ sudo grep electrified.zip /var/log/apache2/www.openrailwaymap.org.access.log.1 | grep JOSM | wc -l
11
```

@DerDakon What do you think?

If we merge this pull request, we have to update the URL in JOSM's lists of map style. JOSM devs will notice the change because they monitor the availability of map styles, presets, WMS/WMTS/tile services.

mmoll:
JFTR, having the redirect documented in the ansible files would also be OK for me.

DerDakon:
We also have the "standard" layer that is called "infrastructure", so names are inconsistent anyway. I would not mind renaming this one, we could just add a redirect in the other direction for the time being then. It was probably my fault to pick a bad name when I created this.

PR #709, opened 2021-02-19
switch CI to GitHub actions

DerDakon:
TravisCI has issues like status reporting sometimes not working, and requires credits meanwhile.


.github/workflows/python.yml
```diff
@@ -0,0 +1,39 @@
...
+        python-version: [3.6, 3.7, 3.8, 3.9]
+
+    steps:
+    - name: checkout
+      uses: actions/checkout@v2
+      with:
+        submodules: true
+
+    - name: deps
+      run: sudo apt-get update && sudo apt-get install libxml2-utils gettext
```
mmoll, line 26:
```suggestion
      run: sudo apt update -qq && sudo apt install -yq libxml2-utils gettext
```

DerDakon: I dropped the update, it works without that.

mmoll (approved):

Nakaner (approved):

PR #711, opened 2021-02-21
define available translations only at one place

DerDakon:
The PHP code already has this list, so just pass it down to JavaScript instead of defining it 3 times again.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -186,6 +186,8 @@
 
 	function urlArgsToParam($checkMobile, $urlbase)
 	{
+		global $langs;
+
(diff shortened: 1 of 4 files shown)


mmoll (approved):
seems to work just fine 👍🏿

DerDakon:
This also fixes the incorrect mapping of the Chinese languages on the website where they are not distinguished, as reported in https://github.com/OpenRailwayMap/OpenRailwayMap/pull/788#issuecomment-1243791434.

Nakaner (approved):

Issue #712, opened 2021-02-24
permalink contains "lang=null" if language was not manually chosen


DerDakon:
- open https://openrailwaymap.org
- scroll/zoom anywhere
- select any layer
- click permalink

The link will contain ```lang=null```. If you manually select a language it will become correct.

PR #713, opened 2021-02-25
fix getPermalinkUrl() to not include lang=null param

mmoll:
closes GH-712

### DIFF
--- a/js/functions.js
+++ b/js/functions.js
@@ -146,7 +146,14 @@ function getPermalinkUrl(style, lang)
 {
 	var url = window.location.origin + window.location.pathname;
 
-	url += '?lang='+((lang) ? lang : params['lang']);
+	url += '?style='+style;
+
+	if (lang === null || lang === undefined) {
+		lang = params['lang'];
+	}
+	if (lang !== null) {
+		url += '&lang='+lang;
+	}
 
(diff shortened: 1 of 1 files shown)


js/functions.js
```diff
@@ -145,8 +145,19 @@ function changeLanguage(lang)
 function getPermalinkUrl(style, lang)
 {
 	var url = window.location.origin + window.location.pathname;
+	var langAppend;
 
-	url += '?lang='+((lang) ? lang : params['lang']);
+	url += '?style='+style;
+
+	if(lang != null) {
```
DerDakon, line 152: what about something like this instead:

```js
if (lang === null) { lang = params['lang']; }
if (lang !== null) { url += &lang="+lang; }
```

mmoll: ✅

Nakaner (changes requested):
It is `lang=undefined` instead of `lang=null` now.

mmoll:
Should be fixed now.

DerDakon:
You can use the same indentation style like the surrounding code…

DerDakon (approved):

Nakaner:
The changes were deployed a few minutes ago.

PR #736, opened 2021-08-22
Track gauge

TrainTraveler:
I have created a new style layer to see the different gauge. 
Please review the files, and feel free to change or improve it.


styles/gauge.mapcss
```diff
@@ -0,0 +1,1165 @@
...
+way[gauge="0'2.5\""].tracks,
+way[gauge=63].tracks,
+way[gauge>63][gauge<88].tracks
+{
+  z-index: 42;
+  color: #006060;
+}
+
+way[gauge="3.5inch"].tracks,
+way[gauge="3_1/2_inch"].tracks,
```
Nakaner, line 393: Underscores and slashes make our MapCSS compiler (run `cd styles && make`) freak out. JOSM would work if we did not use `@import` which requires preprocessing using `awk`. Please get rid of them.

To be honest, I don't understand why you want to support any other way to provide gauge than the pure number in millimeter. They are not frequently used and the map styles should not support broken tagging.

Nakaner (changes requested):

Nakaner:
By the way, please add your new style to the Makefile.

DerDakon:
@Nakaner when merging we need to squash all those commits I think.

styles/gauge.mapcss
TrainTraveler: Done. I removed the inch stuff and other broken tagging. Since it would have to be millimeter anyway, we don't need that here.

TrainTraveler:
> By the way, please add your new style to the Makefile.

Done.

Nakaner (changes requested):
If I add a way in JOSM with `railway=rail`, `usage=main`, `gauge=1522`, I don't see any difference in rendering if I enable this new map style (I load the .zip file in JOSM). Did you test your pull request with JOSM? That's important because JOSM is nowadays the only target platform of the MapCSS styles.

styles/gauge.mapcss
```diff
@@ -0,0 +1,1119 @@
...
+way[railway].tracks
+{
+  text-position: line;
+  text-color: #585858;
+  text-halo-radius: 1;
+  text-halo-color: white;
+  font-family: "Verdana Bold";
+  font-weight: bold;
+  font-size: 11;
+  opacity: #000000;
```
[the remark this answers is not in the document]
TrainTraveler: opacity wasn't explicitly required, I just removed it.
I also tested now the mapcss in JOSM. The line width was missing and therefore not visible if no other style is selected. Sorry for that fault, it should work better now.

Nakaner (approved):

styles/gauge.json
```diff
@@ -0,0 +1,564 @@
...
+			}],
+			"caption": "1450mm"
+		},
+		{
+			"minzoom": 10,
+			"features": [{
+				"type": "LineString",
+				"coordinates": [[10, 50], [80, 50]],
+				"properties": {
+		      			"railway": "rail",
```
DerDakon, line 393: This line has a white space error, there are a few spaces between the tabs. Please remove them.

styles/gauge.mapcss
```diff
@@ -0,0 +1,1134 @@
...
+	default-lines: true;
+}
+
+@import url("common.mapcss");
+
+/* extend common tracks with monorail and miniature */
+
+way|z10-[railway=monorail]
+{
+  set .mono_tracks_present;
```
DerDakon, line 37: Please indent these lines with 2 tabs as the other styles and the lines above also use.

Please also make sure to use Unix line endings, i.e. LF only, as the file currently has CRLF line endings.

DerDakon (changes requested):

styles/gauge.json
TrainTraveler: done

styles/gauge.mapcss
TrainTraveler: done

DerDakon (approved):

PR #743, opened 2021-09-21
Replace awk script resolving MapCSS imports by Python

Nakaner:
For an unknown reason, the GNU AWK script does not work on my Ubuntu 20.04 computer. I don't have enough AWK knowledge to fix this. Therefore, I wrote this Python script doing that task.

### DIFF
--- a/styles/Makefile
+++ b/styles/Makefile
@@ -11,9 +11,9 @@ all: $(JSFILES) $(ZIPFILES)
 %.js: %.mapcss common.mapcss
 	$(PYTHON) ../renderer/mapcss_converter.py --mapcss $^ --icons-path .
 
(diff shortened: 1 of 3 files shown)


DerDakon (approved):
If it works I'm fine with it.

styles/Makefile
```diff
@@ -13,7 +13,7 @@ all: $(JSFILES) $(ZIPFILES)
 
 %.zip: %.mapcss common.mapcss josm-import.awk
```
DerDakon, line 14: the dependency needs to be updated as well

DerDakon (changes requested):

DerDakon (approved):

PR #746, opened 2021-10-27
Update  zh_TW

SteveLz:


Nakaner (approved):
I don't understand the language but it looks fine from technical point of view.

PR #753, opened 2021-11-30
Add property usage:main to lines in electrification legend

### LINKED ISSUE #751: Lines not rendering in electrification legend for zoom < 10
Surprised no one seems to have complained about this yet, but the captions in the legend for the electrification map don't have accompanying colored lines for me for zooms 2 through 9.

Adding the property `usage:main` in the appropriate places in `styles/electrified.json` should bring it in line with the other legends I think.

epimorphic:
Should fix #751, so that the legend for the electrification style isn't a blank box with only text until one zooms in a lot.

### DIFF
--- a/styles/electrified.json
+++ b/styles/electrified.json
@@ -8,7 +8,8 @@
 				"type": "LineString",
 				"coordinates": [[10,50],[80,50]],
 				"properties": {
-					"railway":"rail"
+					"railway":"rail",
+					"usage":"main"
 				}
 			}],
 			"caption": "No information"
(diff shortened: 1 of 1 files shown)


styles/electrified.json
```diff
@@ -20,7 +21,8 @@
 				"coordinates": [[10,50],[80,50]],
 				"properties": {
 					"railway":"rail",
-					"electrified":"no"
+					"electrified":"no",
+					"usage":"main"
```
DerDakon, line 25: Just a small nitpick: if you move the new line right after the ```"railway":"rail"``` then this will only change one line for every hunk but the first one.

epimorphic: Sure, that would be cleaner. Do I make a new PR then?

epimorphic: Should be good now?

DerDakon (approved):

PR #764, opened 2022-02-01
Update German translation

DerDakon:
Also some minor whitespace fixes.

### DIFF
--- a/locales/ca_ES/LC_MESSAGES/messages.po
+++ b/locales/ca_ES/LC_MESSAGES/messages.po
@@ -790,3 +790,141 @@ msgstr ""
 
 msgid "Contrôle de vitesse par balises (KVB)"
 msgstr ""
+
+msgid "1067mm"
+msgstr ""
+
+msgid "1435mm"
+msgstr ""
+
+msgid "1676mm"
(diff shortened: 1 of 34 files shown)


locales/ca_ES/LC_MESSAGES/messages.po
```diff
@@ -790,3 +790,147 @@ msgstr ""
...
+#, fuzzy
+msgid "Broad gauge"
+msgstr "Gàlib"
+
+msgid "1668mm"
+msgstr ""
+
+#, fuzzy
+msgid "Dual gauge"
+msgstr "Gàlib"
```
Nakaner, line 809: Do "broad gauge" and "dual gauge" have the same translation?

Same for other locales, e.g. FI, FR, HU.

Nakaner (changes requested):

locales/ca_ES/LC_MESSAGES/messages.po
DerDakon: msgmerge tried to be clever and added some guesses, I added a commit to stop this.

Nakaner (approved):

PR #767, opened 2022-02-24
Spanish translations improved and completed

ruben2s:
Basically just that. I corrected some translations and added those that were incomplete.


Nakaner (approved):
I don't understand Spanish but I have no technical objections.

PR #771, opened 2022-03-08
Update messages.po

xephos1one:
added a German translation for "Track gauge" when the "(beta)" in "Track gauge (beta)" gets deleted


Nakaner (approved):

DerDakon:
I wonder if there is a usage of that string yet?

DerDakon:
If that string is not in use yet this isn't actually needed. One of us will update the translation file when the "beta" will get remove, we are all native German speakers.

Issue #738, opened 2021-08-29
README.md dead link


bjoernma:
In README.md the link to https://www.openrailwaymap.org/contribute.html is dead (from "how to contribute")

Fyrgeit:
Same bro

PR #780, opened 2022-05-06
Update contribution and maintainer info

DerDakon:
Fixes: #738

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,4 +17,4 @@ We currently do not use any versioning for this repository. That reduces the man
 
 ## Maintainers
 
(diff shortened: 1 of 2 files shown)


rurseekatze (approved):

rurseekatze:
I discovered that I am still mentioned [here](https://github.com/OpenRailwayMap/OpenRailwayMap/blob/master/imprint-de.html#L109) and [here](https://github.com/OpenRailwayMap/OpenRailwayMap/blob/master/imprint-en.html#L108).

Maybe [this](https://github.com/OpenRailwayMap/OpenRailwayMap/blob/master/imprint-de.html#L177) should also be changed from NRW to BW?

PR #788, opened 2022-07-15
Update zh_CN

SteveLz:


DerDakon:
The diff is very large, which comes from the fact that the line endings are using Windows instead of Unix style. Please fix this.

LaoshuBaby:
I'm also a native speaker of zh_CN, I want to translate because I found too much strings haven't been translated, and found this repository. If I haven't seen this PR I will do repetitive work. I roughly review this and think it can be merge. 

Waiting for your opinion! Thank you! @DerDakon

locales/zh_CN/LC_MESSAGES/messages.po
```diff
@@ -1,10 +1,10 @@
...
-# 
-# Translators:
-# FreedSky <qa003qa003@hotmail.com>, 2020
+# OpenRailwayMap Copyright (C) 2018 Alexander Matheisen
+# This program comes with ABSOLUTELY NO WARRANTY.
+# This is free software, and you are welcome to redistribute it under certain conditions.
+# See https://wiki.openstreetmap.org/wiki/OpenRailwayMap for details.
+# 
+# Translators:
+# FreedSky <qa003qa003@hotmail.com>, 2020
```
LaoshuBaby, line 7: Can @SteveLz add his name after this?

DerDakon (approved):

DerDakon:
Deployed.

LaoshuBaby:
> Deployed.

 Looks like in deployed orm site, zh-CN's string still have error? It haven't use this PR's new string.

图片

DerDakon:
> > Deployed.
> 
> Looks like in deployed orm site, zh-CN's string still have error? It haven't use this PR's new string.

I forgot to call make, should be fixed now.

SteveLz:
Unfortunately, the picture shows another problem. The circled place does not call the string from zh_CN but from zh_TW.

DerDakon:
This happens because the mapping of the Chinese languages is wrong in the current code: https://github.com/OpenRailwayMap/OpenRailwayMap/blob/master/js/bitmap-map.js#L43

This should be solved by #711 as well, even if that fix was not intended for that change. It has been fixed in the code currently deployed locally in a different way for the moment, so this should now work.

PR #789, opened 2022-07-16
Update zh_TW

SteveLz:


Nakaner (approved):
I can't comment on the language because I do not understand it. However, I have no objections from technical point of view. I assume that removing the space between the number and the unit complies with your local spelling rules.

DerDakon:
I'll merge this when the "space thing" has been answered.

DerDakon:
@SteveLz please comment on the "space between number and unit" thing that Nakaner mentioned above.

SteveLz:
There is no uniform standard for whether to add spaces between numbers and Chinese characters in Chinese. One argument is that the distance between them should be achieved by typesetting rather than adding spaces, generally keeping a quarter full-width character distance. In practice, it seems that there are more without spaces than with spaces.

PR #793, opened 2022-09-03
Speed limit sign questionmark

rurseekatze:
A "?" is a valid value for unknown speed limit values, not only for light signals, but also for signs.

Currently the validator throws a false positive for https://www.openstreetmap.org/node/6067379690.

Instead of an error, just throw a warning to support the mapper to find incomplete signal tagging.


DerDakon:
I would downgrade this even to "throwOther" as the tagging is in no way wrong, it's just about missing data.

rurseekatze:
> I would downgrade this even to "throwOther" as the tagging is in no way wrong, it's just about missing data.

I was not sure whether "throwWarning" or "throwOther" is the better category, but you are right. I changed it.

DerDakon (approved):

PR #795, opened 2022-09-17
Update messages.po

dragonmoonpt:
Update some terms in Portuguese


DerDakon (approved):
Not that I understand any of this, but it looks sane.

PR #797, opened 2022-09-28
Add inner/outer curved switches to josm preset

bjoernma:
Sync preset to documentation.


DerDakon (approved):
Looks good, although I had to search for the exact term in the wiki to find the place you thought about. https://wiki.openstreetmap.org/wiki/DE:Tag:railway%3Dswitch doesn't have it.

bjoernma:
I had understood the wiki in a form that nearly everything related to ORM is in the Tagging or Tagging_in_Germany pages. If that is not entirely the case, we should maybe think about better sync methods for that and less redundance?
Thanks :)

Issue #763, opened 2022-01-16
rename "de.xml" JOSM template


DerDakon:
The comment at the top says it's about infrastructure in Germany and Austria, but it can be used worldwide. I think it's less confusing if we rename it.

bjoernma:
The issue can be closed (fixed by merging #801).

PR #801, opened 2022-10-22
rename de.xml to infrastructure.xml; adjust entries

bjoernma:
Yet missing to have it globally usable:
- add all the other train control systems, maybe as a combo box
- Is there a better solution for the derail/buffer_stop-signals? Is a global list of possibilities to overwhelming? 
- Are there radio systems missing? Add them.

For some naming it's more focused on german/austrian systems yet, but should be globally viable.
Should solve Issue #763

### DIFF
--- a/josm-presets/Makefile
+++ b/josm-presets/Makefile
@@ -2,7 +2,7 @@ SHELL = /bin/sh -e
 
 .PHONY: check clean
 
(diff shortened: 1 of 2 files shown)


Nakaner (changes requested):
Line 79 needs to be changed as well:

`	<group name="OpenRailwayMap Infrastructure DE+AT" de.name="OpenRailwayMap Infrastruktur DE+AT" icon="signals.png">`

Nakaner (approved):

Issue #811, opened 2022-12-29
Update Taginfo Projects


danieldegroot2:
Show the OpenStreetMap community what [data](https://wiki.openstreetmap.org/wiki/Tags) you are using. Update your project on the [list](https://taginfo.openstreetmap.org/projects) of [Taginfo Projects](https://github.com/taginfo/taginfo-projects).

Was: #60

hiddewie:
I saw this issue by accident when I was scrolling the issue list here. I've made a PR to document many of the used tags in https://github.com/OpenRailwayMap/OpenRailwayMap/pull/812.

PR #812, opened 2022-12-29
Document many used tags for the TagsInfo project

hiddewie:
Related to https://github.com/OpenRailwayMap/OpenRailwayMap/issues/811

This PR documents many of the used tags by the OpenRailwayMap project.

I used a command to query all tags from the CartoCSS `.mml` files, and added some more used tags manually to the list, also inspired by the osmium command https://github.com/OpenRailwayMap/OpenRailwayMap-CartoCSS/blob/master/SETUP.md#load-osm-data-into-the-database.

Format is described on https://wiki.openstreetmap.org/wiki/Taginfo/Projects.

See projects on https://taginfo.openstreetmap.org/projects.

### DIFF
--- a/taginfo/taginfo.json
+++ b/taginfo/taginfo.json
@@ -1,8 +1,8 @@
 {
 	"data_format": 1,
 	"data_url": "https://www.openrailwaymap.org/taginfo/taginfo.json",
-	"project":
-	{
+	"data_updated": "20221229T183000Z",
+	"project": {
 		"name": "OpenRailwayMap",
 		"description": "An OpenStreetMap-based project for creating a map of the world's railway infrastructure",
 		"project_url": "https://www.openrailwaymap.org/",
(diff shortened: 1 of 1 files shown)


DerDakon:
I think we should aim a little higher, like having that command as a make rule or at least a script that we can run every now and then to update the list. But… this removes the correct "buffer stop" now, which I don't like. Also we do not use the route or public_transport information, we actually filter it out from our database to not interfere with the rendering. At least we did in the past.

DerDakon:
And we should update the maintainer name here as well, right? @Nakaner

hiddewie:
> I think we should aim a little higher, like having that command as a make rule or at least a script that we can run every now and then to update the list. 

Yes that would be nice. Several problems that I could find yesterday when I compiled this set of tags with values:
- I parsed the `.mml` files in the `OpenRailwayMap-CartoCSS` which may be on a different directory on the filesystem, so cannot be run from a script in this repository.
- The osmose filter was a manually added list. The osmium tag filter includes the types of objects (also added to the taginfo file), but this is not easy to parse automatically. 
- I manually added some more tags to the taginfo file that were indirectly used in the SQL queries, but are not of the form `tags->'...'`.

> But… this removes the correct "buffer stop" now, which I don't like.

Currently, 
```
  {
    "key": "railway"
  },
```
is included, which includes every value of `railway`, including buffer_stop. I think this is fair for a railway-oriented map.

> Also we do not use the route or public_transport information, we actually filter it out from our database to not interfere with the rendering. At least we did in the past.

I took those tags from the osmium tag filter in the `OpenRailwayMap-CartoCSS` repository:
```
osmium tags-filter -o filtered.osm.pbf input.osm.pbf nwr/railway nwr/disused:railway nwr/abandoned:railway nwr/razed:railway nwr/construction:railway nwr/proposed:railway n/public_transport=stop_position nwr/public_transport=platform r/route=train r/route=tram r/route=light_rail r/route=subway
```

So I assumed that the `route` and `public_transport` are used for something, otherwise the objects would be left out of the import. I can remove those tags if they are indeed not used.

Issue #824, opened 2023-03-18
Add rendering for train protection in Sweden


tomastomas75:
Looking at the map style "signalling", all railways in sweden are marked as "no information available" (except for Bottniabanan which is tagged railway:etcs=yes). Which is not true. The information IS available.

Most of swedish railways do have the ATC system (automatic train control), since the 1980's and are also tagged 

railway:atc=yes
or
railway:atc = no
(where "no" means "no protection system", not "no information available")

This is also, as I understand, tagged in the same way as german railways which have:

railway:pzb=yes
railway:lzb=yes

..and others. 

I guess it wouldn't be TOO much work to add a legend for swedish ATC (and norwegian, which is identical). Just look for the tag "railway:atc=yes" end give it a nice colour.

bjoernma:
Does anybody know whether there is already a (tagging) distinction between the US / japanese / norwegian ATC system as they sound to be (technically) different in wikipedia. 
I think that will be an even bigger problem for the "no protection" case.
If there would be a distinction in tagging, my next question for sweden/norway would be about an complete(!) list of possible train control systems in those two countries. 
I think a country based rendering is not (yet?) possible?

tomastomas75:
I don't know anything about US or Japan, but in Sweden there are only ATC or ETCS, or no train protection* at all.

Norway has the same ATC system as Sweden, and I dont THINK there is any other systems there, but I'll check. 

*) Manual systems, as "calling the other end of the line to receive authorization" counts as "no protection".

DerDakon:
ATC should now be rendered.

tomastomas75:
Excellent @DerDakon, thank you!

PR #825, opened 2023-03-24
Add ATC to legend

bjoernma:
as suggested in #824 added train protection system ATC to the legend. 
I expect it to be used as well for non-european systems, so I used a nonexisting colour. 
The "no-train-protection" case is at the moment the try to distinguish between european and non-european system so I expect that etcs-tags only occur in europe (so missing out the [!"railway:etcs"] case for it). 
Edit: just saw that etcs is also used in various non-european countries. So what's better? Remove the no-protection condition again or live with maybe false-negatives (there could be train protection even when "no" is shown on the map)?

### DIFF
--- a/styles/signals.json
+++ b/styles/signals.json
@@ -106,6 +106,19 @@
 				"coordinates": [[10,50],[80,50]],
 				"properties": {
 					"railway":"rail",
+					"railway:atc":"yes",
+					"usage":"main"
+				}
+			}],
+			"caption": "ATC - Automatic train control"
+		},
+		{
+			"minzoom": 2,
+			"features": [{
+				"type": "LineString",
+				"coordinates": [[10,50],[80,50]],
+				"properties": {
+					"railway":"rail",
 					"railway:lzb":"yes",
 					"usage":"main"
 				}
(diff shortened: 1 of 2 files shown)


bjoernma:
Short answer: atm it has to be "german"=no or (atb AND etcs)=no (and I added something like "(atc AND etcs)=no" and am not sure about problems in USA/Japan 

As it's a bit difficult case (atc used in various countries..), I decided to use atc=no and etcs=no for the moment. I don't know how to allow for a more differentiated case that it's possible to have only a per-country rule. Would it be an option to use another tagging to differentiate differently working ATC systems as of their country/version? (and as it is a different system it should not be tagging-for-the-renderer I think).
It seems like the whole train protection tagging system is a bit off and not designed for overlaps.

davide84:
Hi! Regarding the "no" conditions, it's getting more and more complicated as new systems are added.
There is already a [proposal](https://wiki.openstreetmap.org/wiki/Proposed_features/railway:train_protection) to change the tagging scheme and IMHO it would completely solve this issue. I would be happy if you also comment in that discussion and push for its approval :)

bjoernma:
Oh yes, that sounds very necessary. Thanks!

DerDakon (approved):

PR #828, opened 2023-04-03
MOD: changed ZSI 127 color to brown to resolve conflict

davide84:
The recently added ZSI 127 has a violet color in the legend. It matches the color of the scandinavian ATC; since the rendering part of ZSI 127 in the Carto repository has not been added yet, I tought of changing the color now to avoid bigger fixes later.

I picked up brown because ZSI 127 is only meant to be used in the narrow-gauge network in Switzerland, which will be 100% blue because of ETCS, so I mainly wanted to avoid blue (and Italy's pink). Being a German-speaking region I thought of picking up a color that would reflect this inheritance.

### DIFF
--- a/styles/signals.json
+++ b/styles/signals.json
@@ -103,6 +103,19 @@
 			"minzoom": 2,
 			"features": [{
 				"type": "LineString",
+				"coordinates": [[10, 50],[80,50]],
+				"properties": {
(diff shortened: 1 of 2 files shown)


davide84:
Screenshot from 2023-04-03 12-40-58

PR #839, opened 2023-07-13
Update Hungarian translation

BathoryPeter:


locales/hu_HU/LC_MESSAGES/messages.po
```diff
@@ -716,212 +717,212 @@ msgid "Ks-Vorsignal (normal, Wiederholer, verkürzter Bremswegabstand)"
 msgstr ""
 
 msgid "Bü 4 Pfeiftafel (normal, für durchfahrende Züge)"
-msgstr ""
+msgstr "Bü 5 kürtölni tábla (normál, áthaladó vonatoknak)"
```
Nakaner, line 720: Shouldn't this be Bü 4 even in Hungarian? ;-)  Please check if the remaining Hungarian translation is not a copy and past mistake as well.

Nakaner (changes requested):

PR #841, opened 2023-07-28
Validate against razed railways with existing properties

Famlam:
When a railway has completely disappeared from the earth (i.e. `railway=razed`), it cannot have properties like a voltage or a gauge width anymore. These properties only applied to the object when it still existed. Hence the only logical way to tag these is to have them with the appropriate lifecycle prefix.

This PR will add a warning for such cases.

(I didn't add them to `railway=abandoned` as (for example) the gauge may (hypothetically) be verifiable based on remaining traces like sleepers)

### DIFF
--- a/validator/openrailwaymap.validator.mapcss
+++ b/validator/openrailwaymap.validator.mapcss
@@ -820,3 +820,33 @@ node["railway:signal:speed_limit:speed"~="?"]
 	assertNoMatch: "node railway=signal railway:signal:speed_limit:speed=120";
 	assertNoMatch: "node railway=signal railway:signal:speed_limit:speed=\"80;90\"";
 }
+
+/* Razed railways with active properties */
+way[railway=razed][electrified],
+way[railway=razed][frequency],
+way[railway=razed][gauge],
+way[railway=razed][service],
+way[railway=razed][tracks],
+way[railway=razed][usage],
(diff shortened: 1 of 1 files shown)


DerDakon:
I have not tested it yet, but the idea sounds good.

Nakaner:
It is fine from programming point of view. However, I wonder whether we really need to replace `gauge=*` by `razed:gauge=*`. For example `razed:gauge=*` is used 423 times but `railway=razed` + `gauge=*` is used 16816 times. We use these prefixed keys for electrification and gauge if that is changed while a line is still in use (e.g. a non-electrified track is going to be electrified). However, in the case of razed railways, it is obvious that the tag refers to the historic attributes. Therefore, I am tempted to reject this pull request. But I am happy to hear other opinions (and am open to discuss this on community.openstreetmap.org).

Famlam:
I was primarily thinking about cases like these:
https://www.openstreetmap.org/way/759730445
https://www.openstreetmap.org/way/169576296

Usually such cases are a compromise between keeping the historic railway information ("the highway follows the trace of a former railway") without keeping non-existent items as separate objects in OSM.

Here the "active" tag (`highway` in both examples) now has an "active" gauge, while that gauge should refer to the razed railway: the highway itself (hopefully) doesn't have a gauge.

PR #851, opened 2023-08-29
Use preferred tile.openstreetmap.org URLs

freaktechnik:
See openstreetmap/operations#737


Nakaner:
Thank you for providing the patch. I deployed this patch a few minutes ago.

PR #855, opened 2023-10-06
Add german/english translations

bjoernma:
A bunch of german/english translations were missing and were added, especially for german signal systems.


Nakaner (changes requested):
I think that it should be a translation, not a translation with explanation. The translations of some Hp signals are way longer than their German equivalents.

PR #856, opened 2023-10-07
Improved track gauge legend

bjoernma:
The color in the legend of track gauges is not nicely visible due to the labels, therefore elongate the line.
Squash similar parts with replace function for better readability.


Nakaner (changes requested):
Looks good to apart from the single empty line (see comment).

styles/gauge.json
```diff
@@ -481,84 +352,31 @@
...
-				"coordinates": [[10, 50], [80, 50]],
-				"properties": {
-					"railway": "miniature",
-					"gauge": "190"
-				}
-			}],
-			"caption": "190mm"
-		},
-		{
+		},{
```
DerDakon, line 361: linebreak

DerDakon (changes requested):
Thanks for the cleanup. You will get bonus points when you include an image of the updated rendering ;)

styles/gauge.json
bjoernma: no bonus I guess... I just cant get the complete chain running and end up with uninterpreted html code.

PR #862, opened 2024-01-03
josm preset german ESO substitute signals

bjoernma:
For most types of main and combined signals it was not possible to have "no substitute signal" in the josm preset. This PR will add those. Additionally some names were put as substitution signals. This is made consistent to substitute signal. 

The names for the Zs7 (Vorsichtsignal) are not consistent in english yet as I think that would need some more discussion and specialized english knowledge. At the moment it's a mix of cautioness signal and cautiousness signal. Are we sure that it's not only a type of permissive signal? One could as well discuss about substitute signal as on some places I heard more about a "call on signal" instead.

### DIFF
--- a/josm-presets/de-signals-eso.xml
+++ b/josm-presets/de-signals-eso.xml
@@ -159,9 +159,9 @@ See https://wiki.openstreetmap.org/wiki/OpenRailwayMap for details.
 						text="Substitute signal (multi-selection)"
 						de.text="Ersatzsignal (Mehrfachauswahl)"
 						delimiter=";"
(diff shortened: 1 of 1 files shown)


Nakaner (approved):
Thank you for the contribution.

It is now possible to add `railway:signal:main:substitute_signal=DE-ESO:dr:zs1;no` but there is no option to have a simple UI where this is not possible.

If you want, you can add a new check to the validation rules for this mistake.

bjoernma:
Thanks, I will think about the validator in another PR if I get it running.

Issue #689, opened 2020-07-26
Specific validator rules for Grenzzeichen (DE-ESO:ra12)


bjoernma:
Atm, the signal for fouling points at switches always shows errors JOSM style/preset (yellow image with prohibition sign and exclamation mark). It should be changed in a way that the missing of the tags railway:signal:direction and railway:signal:position doesn't show an error as they are not useful for point-like signals. 
Furthermore it seems that this signal is often placed in the real position and not belonging to a track. Check for that?

PR #864, opened 2024-01-06
Render fouling points in josm-style

bjoernma:
As described in #689 fouling points in germany were not rendered, but (as they are often not tagged with railway:signal:direction) often as a "deprecated" symbol. 

- Adds the rendering itself for ESO, BOStrab and PKP fouling points (they all look the same)
- Doesn't care about direction tag in this special case


bjoernma:
Auswahl_072

Nakaner (approved):

PR #866, opened 2024-02-04
Update locales

DerDakon:
Refresh template, and the different translation files from the template.


Nakaner (changes requested):
This pull request added a few new strings which look new because they replace older (removed) strings. This leads to duplication in the locales:

* "LZB - Linienzugbeeinflussung" vs. "Linienzugbeeinflussung" (old string replaced by new one)
* "PZB - Punktförmige Zugbeeinflussung" vs. "Punktförmige Zugbeeinflussung" (old string replaced by new one)
* "no train protection" vs "No train protection" (spelling change)
* "ETCS - European Train Control System" vs. "ETCS" (old string replaced by new one)

PR #872, opened 2024-02-20
Update startposition.js

RailMapper:
Add marker if if lat, lon, zoom and optionally mapstyle are given in the url. Example: https://www.openrailwaymap.org/?style=maxspeed&lat=50.134636814051056&lon=14.403119087219237&zoom=15

### DIFF
--- a/js/startposition.js
+++ b/js/startposition.js
@@ -45,6 +45,13 @@ function Startposition(map)
 		else
 			self.setPosition();
 	};
+
+	// add marker
+	var marker = L.marker([params['lat'], params['lon']], {
+		src: "https://openrailwaymap.org/js/images/marker-icon.png"
+	});
+	marker.addTo(this.map);
+	
 	// permalink given
 	if (params['lat'] && params['lon'])
 	{


js/startposition.js
```diff
@@ -45,6 +45,13 @@ function Startposition(map)
 		else
 			self.setPosition();
 	};
+
+	// add marker
+	var marker = L.marker([params['lat'], params['lon']], {
+		src: "https://openrailwaymap.org/js/images/marker-icon.png"
+	});
+	marker.addTo(this.map);
+
```
Nakaner, line 54: This fails if the parameters are not part of the query string (unset) and lead to a JavaScript error. If we accept this pull request, your inserted sections needs to moved behind the call of `this.map.setView`.

Nakaner (changes requested):
While support for a marker is good, it changes behaviour users have been used to for more than ten years now. `lat`and `lon` are used to share a view. I would accept this pull request if `marker=true` in query string were the condition for displaying a marker.

js/startposition.js
RailMapper: The insertion of the proposed changes in startposition.js I did at the wrong position, sorry. Will be changed immediately.

But in my opinion the additional <code class="notranslate">marker=true</code> has to be assigned on the serverside as part of the query string to the dictionary <code class="notranslate">params</code>. Is it right or have I to do further changes?

PR #876, opened 2024-03-24
Add BOStrab main signals to signals.json

MrKrisKrisu:
> Could you please update the [map key](https://github.com/OpenRailwayMap/OpenRailwayMap/blob/master/styles/signals.json#L421)?

see https://github.com/OpenRailwayMap/OpenRailwayMap-CartoCSS/pull/124#issuecomment-2016563724

Sure! I'm a little bit unsure about this change. Is this the way, the change is required?
(@Nakaner)

### DIFF
--- a/styles/signals.json
+++ b/styles/signals.json
@@ -419,19 +419,41 @@
 		},
 		{
 			"minzoom": 14,
+			"maxzoom": 14,
 			"features": [
 				{
 				"type": "Point",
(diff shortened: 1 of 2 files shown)


Nakaner:
I added the missing styling to the MapCSS style and changed the map key to make it easier to understand. I decided against updating the locale files because there are already open pull requests for it.

@DerDakon Any comments?

PR #892, opened 2024-08-28
Update python versions

bjoernma:
The workflow works yet, but python 3.7 has reached the end of updates last year. Added python 3.10 as that seems to be the standard version in ubuntu 22.04.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.8, 3.9, "3.10"]
 
     steps:
     - name: checkout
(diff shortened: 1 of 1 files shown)


Nakaner:
The build with Python 3.10 fails. I suspect that it is cause by the old version of PIL we require. I succeded to build the MapCSS styles on Ubuntu Jammy and Noble with python3-pil from APT (9.0.2 on Jammy, 10.2.0 on Noble). I am currently trying a newer version of PIL in [a seperate branch](https://github.com/OpenRailwayMap/OpenRailwayMap/commit/751b3ea8f7e64823a67460ec276924ddda1a86f8). Let's see if the build succeeds.

Nakaner:
It worked and I pushed the commit onto your branch.

Nakaner (approved):

bjoernma:
> It worked and I pushed the commit onto your branch.

Best thanks!

PR #898, opened 2024-11-16
add: Turkish translate added for blank translations.

ramazansancar:
upd: Makefile script add install polib dependency. And update script.
upd: run update_po script. All Locales files updated.


ramazansancar:
I saw explanations in German in many of the sources. Is there an action plan for them?

PR #905, opened 2025-05-23
ADD: British AWS signalling in legend

prophile:
There are data for TPWS as well, for now let's disregard that though.


Nakaner:
Thank you for your contribution. I applied the changes to the MapCSS style and the JOSM preset. However, the map rendering is [implemened in CartoCSS](https://github.com/OpenRailwayMap/OpenRailwayMap-CartoCSS). As long as that map style does not support `railway:aws`, an entry in the map key will not make sense. Therefore, I will keep this pull request unmerged but two thirds of your contribution are already merged.

PR #922, opened 2026-02-07
Update zh-TW l10n

petercpg:
# Changes
1. Completed translation
2. Corrected incorrect translation and wordings
3. Corrected translation based on OSM tagging scheme and Taiwanese locale usage
4. Reverted German- and Finnish- signal system names to original as there is no good translation for these

### DIFF
--- a/locales/zh_TW/LC_MESSAGES/messages.po
+++ b/locales/zh_TW/LC_MESSAGES/messages.po
@@ -11,16 +11,16 @@ msgstr ""
 "Project-Id-Version: OpenRailwayMap\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2013-03-08 12:29+0100\n"
-"PO-Revision-Date: 2017-09-22 15:50+0000\n"
-"Last-Translator: Supaplex <bejokeup@gmail.com>\n"
-"Language-Team: Chinese (Taiwan) (https://www.transifex.com/rurseekatze/openrailwaymap/language/"
-"zh_TW/)\n"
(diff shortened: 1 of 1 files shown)


locales/zh_TW/LC_MESSAGES/messages.po
```diff
@@ -373,7 +376,7 @@ msgid "Factory connected to railroad line"
 msgstr "連接到鐵路的工廠"
 
 msgid "Rack"
-msgstr "裝卸棧橋"
+msgstr "齒軌"
```
Nakaner, line 379: This change reverts commit c1e890ff8fb4d8e4dec4c22e3a84bf8ae779f13d by @SteveLz 

I can't decide because I don't speak the language. Maybe one of you misunderstood the context? Could you please find an agreement who of you is (more) correct?

petercpg: Can you confirm the source of this string is https://github.com/OpenRailwayMap/OpenRailwayMap/blob/master/josm-presets/infrastructure.xml#L401 ?  Actually, I can not find it from the ORM web.

Based on the JOSM file above I can refer the tag as [rack=*](https://wiki.openstreetmap.org/wiki/Key:rack) and it definetely is [齒軌鐵路](https://zh.wikipedia.org/wiki/%E9%BD%92%E8%BB%8C%E9%90%B5%E8%B7%AF) if you switch to English Wikipedia. 

裝卸棧橋 is more referered as "loading/unloading trestle", the term is used in China but hardly used in Taiwan.

Issue #917, opened 2025-10-30
False positive in validator and italian speed signs


giopera:
Hi everyone,
After working on the Italian railway tagging scheme I've noticed that issues like this (https://osmose.openstreetmap.fr/en/issue/1b5aedc2-517d-6152-8981-4c95727b97e5) are against our tagging schemes since the speed limits in Italy have three, two or one ranks (Here for the tagging scheme https://wiki.openstreetmap.org/wiki/OpenRailwayMap/Signals_in_Italy#Speed_Limits), so is there any way to change this type of behavior?

Nakaner:
The bug is located here: https://github.com/OpenRailwayMap/OpenRailwayMap/blob/master/validator/openrailwaymap.validator.mapcss#L169

Excerpt (shortened):

```css
node[railway=signal]["railway:signal:speed_limit:form"=sign]["railway:signal:speed_limit:speed"=~/;/],
node[railway=signal]["railway:signal:speed_limit_distant:form"=sign]["railway:signal:speed_limit_distant:speed"=~/;/],
{
	throwError: "A sign cannot have different states.";
	assertMatch: "node railway=signal railway:signal:main:states=hp0;hp1 railway:signal:main:form=sign";
	assertMatch: "node railway=signal railway:signal:speed_limit:form=sign railway:signal:speed_limit:speed=80;90";
	assertNoMatch: "node railway=signal railway:signal:main:states=hp0;hp1 railway:signal:main:form=semaphore";
	assertNoMatch: "node railway=signal railway:signal:main:states=hp0;hp1 railway:signal:main:form=light";
	assertNoMatch: "node railway=signal railway:signal:speed_limit:form=sign railway:signal:speed_limit:speed=80";
}

```

You have to add an additional condition selector (the square brakets) which excludes those Italian signals using a [negated regular expression match](https://josm.openstreetmap.de/wiki/Help/Styles/MapCSSImplementation#Conditionselector). Do you want to create the pull request yourself? Pleaes don't forget to add a test (`assertNoMatch`).

Famlam:
@Nakaner there was a small typo in the fix, see the PR I just uploaded. Not sure if this issue needs to be reopened?
In short, the assert tested for `railway:signal:speed_limit_distant` (which is probably correct) but the rule tested for `railway:signal:speed:distant` instead

PR #925, opened 2026-03-28
Fix mapcss mistakes in validator

Famlam:
Fixes:
- Rule and assertNoMatch didn't match for Italian exception - see https://github.com/OpenRailwayMap/OpenRailwayMap/issues/917 and https://github.com/OpenRailwayMap/OpenRailwayMap/pull/918 )
- Removes duplicate asserts in the same rule (resulting in `E: Cannot add MapCSS rule: Cannot add instruction assertNoMatch: way railway=rail operator=DB!` in JOSM)


DerDakon (approved):

Famlam:
p.s. I assumed I didn't have to update the version number (2017-01-09) as it hasn't been updated for years
(This version number is also shown here: https://josm.openstreetmap.de/wiki/Rules )

